PR #40, opened 2017-05-09
move linters to guides [Delivers #50hc697n]

balvig:


linters/rubocop.yml
```diff
@@ -0,0 +1,313 @@
+AllCops:
+  TargetRubyVersion: 2.3
+  Exclude:
+    - config/unicorn/**/*
+    - db/schema.rb
+    - db_chat/schema.rb
+    - vendor/**/*
+    - lib/tasks/dodocook/**/*
```
Knack: We should move this line (and `db_chat/schema.rb` and maybe the `unicorn` line) to the internal `.rubocop.yml` in our global-web project.

linters/README.md
```diff
@@ -0,0 +1,12 @@
...
+## Usage
+
+Add to the following files:
+
+#### .rubocop.yml
+
+```yml
+inherit_from:
+  - https://raw.githubusercontent.com/cookpad/guides/master/linters/rubocop.yml
+```
```
Knack: I would add:

---

#### .gitignore

Ignore `.rubocop-http*` files which are cached files.

```
.rubocop-http*
```

linters/rubocop.yml
balvig: Good point 👍

linters/README.md
balvig: Ah, maybe this belongs in your global gitconfig instead? I don't see those files personally (could be IDE related).

linters/rubocop.yml
balvig: https://github.com/cookpad/guides/pull/40/commits/952ea91fd3a100485670ecfaa905dfd2e5f8b2e1

Knack: 🤔 Maybe the whole `AllCops` block should be specified in each project? We are assuming here a specific Ruby version and a project structure.

Knack: See this repo as an example: https://github.com/pixta-dev/pixta-rubocop

balvig: Also a good point 👍  😅

balvig: https://github.com/cookpad/guides/pull/40/commits/3e5607cec401473f71aaad5fe10f7866ba48005d

linters/README.md
Knack: Maybe it's related to _your_ IDE :-). I see it as a untracked file when running `git status` from the command line. I guess every developer will have this issue, so it's ok to write some documentation explaining it. They can later decide where to ignore it.

balvig: > I guess every developer will have this issue

lol, what I mean is, this file does not get generated on my system (so in other words not an issue for _every_ developer 😉 ). 

I tend to add things that I need in every project to my own `.gitignore_global` file to not have to add it every time. Especially if it is environment-dependent (ie `.DS_Store` etc).

Knack: > maybe I need to do the external file thing first though
🙄

PR #43, opened 2017-06-08
Exclude routes from block rule

balvig:
_wip__investigate_slow_spring__delivers__jklws2cz__by_balvig_ _pull_request__5371_ _cookpad_global-web

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,7 @@
 Metrics/BlockLength:
   Exclude:
     - spec/**/*
+    - config/routes*
 
 Metrics/LineLength:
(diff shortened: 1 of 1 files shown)


Knack:
FYI: This is what I had in newsletters:
```yaml
Metrics/BlockLength:
  Exclude:
    - 'spec/**/*.rb'
    - 'config/**/*.rb'
    - 'lib/tasks/**/*.rake'
    - 'Gemfile'
```

balvig:
that looks good, should we maybe apply that to guides? (feel free to open a PR! 🙏 )

PR #44, opened 2017-06-14
Disable SkipsModelValidations cop

balvig:
Should be a case-by-case basis for this I think? If we need to mess with the DB directly it should be our prerogative 😅 

- http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/SkipsModelValidations
- https://github.com/cookpad/global-web/pull/5412#discussion_r121676698


Knack:
:+1:

PR #45, opened 2017-06-14
Exclude HttpPositionalArguments cop for API tests

balvig:
https://github.com/cookpad/global-web/pull/5422#discussion_r121855642


Knack:
:-1:  https://github.com/cookpad/global-web/pull/5422#pullrequestreview-43959651

balvig:
lol, indeed although the hound is still wrong (the comment is regarding a different method found in plain request specs), maybe we can (ab)use that to catch the kind of mistake you found 🎐

PR #47, opened 2017-06-16
Document scope preference

balvig:


best-practices/samples/rails/5.rb
```diff
@@ -0,0 +1,31 @@
+## Bad
+class Recipe
+  scope :published_on, -> (date) { where(published_on: date) }
+end
+
+## Good
+class Recipe
+  def published_on(date)
+    self.where(published_on: date)
```
JuanitoFatas: Did you mean 🤔 ?

```ruby
def self.published_on(date)
  where(published_on: date)
end
```

balvig: I did indeed 😓  https://github.com/cookpad/guides/pull/47/commits/b4df3317597ccdd1b9dce4043928ff4024125f5b

Thank you!

best-practices/README.md
```diff
@@ -11,6 +11,7 @@
 - Prefer `Date.current` over `Date.today`
 - [Prefer extracting private methods over setting variables inside methods](samples/rails/1.rb)
 - [Extract long chains of unnamed conditions into named scopes or class methods](samples/rails/2.rb)
+- [Prefer class method over scope if the scope takes an argument or spans multiple lines](samples/rails/5.rb)
```
JuanitoFatas: What do you think about we prefer class method names over scopes in general:

```
Prefer `self.method_name` over `scope :method_name` DSL
```

so no need to remember if it takes an argument or spans into multiple lines?

balvig: It might be something to keep an eye on the future! Currently this is just jotting down our current practice, rather than making a new suggestion :)

I guess I like the terseness of scope methods for something simple like:

```ruby
scope :approved, -> { where.not(approved_at: nil) }
```

PR #48, opened 2017-06-16
Exclude special ruby files from BlockLength rule [Delivers #1prpIT5g]

Knack:
Trello: https://trello.com/c/1prpIT5g

In addition to specs, this change excludes special Ruby files that usually don't define Ruby classes with methods, but are mostly based on blocks: configuration files, initializers, rake tasks, etc.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,9 @@
 Metrics/BlockLength:
   Exclude:
-    - spec/**/*
-    - config/routes*
+    - spec/**/*.rb
+    - config/**/*.rb
+    - lib/tasks/**/*.rake
+    - Gemfile
 
 Metrics/LineLength:
   AllowHeredoc: true


sebasoga:
@knack are there any violations that have been triggered lately on the files you are excluding? We probably don't want to exclude stuff unless its annoying us.

Knack:
This is the number of offenses with the current configuration (it excludes specs and routes files):

Project | web | search | newsletters | payments
--------|-----|--------|-------------|----------
app | 0 | 2 | 0 | 0
lib/tasks | 18 | 17 | 5 | 0
config | 4 | 10 | 3 | 0
Gemfile | 0 | 0 | 0 | 0

I guess we don't need to exclude `Gemfile`

> We probably don't want to exclude stuff unless its annoying us.

Of course, it depends on your _level of annoyance_ 🙂 In the newsletters project, for example, the number of RuboCop offenses is 0, so having just a new offense is really annoying. (*) 
But if you have hundreds of offenses you probably won't care too much about these. 

(* there's a trick: we keep a `.rubocop-todo.yml` file to exclude a few things that aren't easy to fix like, for example, this cop)

PR #50, opened 2017-08-02
Do not push big files to GitHub

davidstosik:
See title. :point_up:

I'd even like to suggest changing the wording in operations.md, from

> Don't commit large (15MB+) in the git repo itself!

to

> Avoid pushing temporary data files or big binary files to GitHub

If the file is a temporary data file (eg. CSV used in one-offs), or a big binary file (eg. GeoIP database), then it should not inflate the repository.

I even think it would be nice if we could prevent any big file (and/or extension?) to be pushed, before it happens, because if a file is pushed to the repository, then it is too late, and a [tedious repo cleaning](cookpad/global-web/issues/4777) is necessary to get rid of its weight.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -54,3 +54,4 @@
 
 - When syncing with master, if the branch is under PR review or being collaborated on,
   use `merge` instead of `rebase` to avoid losing relevant context/breaking other local branches
+- Avoid pushing temporary data files or big binary files to GitHub.


Knack:
> I even think it would be nice if we could prevent any big file (and/or extension?) to be pushed

@davidstosik I recently found [git-lfs](https://git-lfs.github.com/). Maybe it could be a solution.

sorah:
Yes, git-lfs is a one solution. But it requires additional setup to users and servers, so we're currently on S3 way (which @davidstosik is mentioning and I've implemented.) I think we should consider git-lfs again if we have a large file which receives a change frequently (i.e. having a need of version control).

davidstosik:
@sorah Do you know of any system (commit/push hooks?) that would allow us to block a big files before it makes it into the GitHub repository?

sorah:
GitHub.com doesn't allow any hook to prevent incoming pushes.

In Enterprise instances, site admins allowed to configure maximum file size...

davidstosik:
It's great to know there's a way! 👍 
I'm sad to read it's not available on GitHub.com. 😢

sorah:
Making a precaution in pull request, using commit status...? 🤔

best-practices/README.md
```diff
@@ -54,3 +54,6 @@
 
 - When syncing with master, if the branch is under PR review or being collaborated on,
   use `merge` instead of `rebase` to avoid losing relevant context/breaking other local branches
+- Avoid pushing temporary data files or big binary files to GitHub ; they inflate
```
balvig: @davidstosik what do you think of shortening this to only say "Avoid pushing temporary data files or big binary files to GitHub" (leave out the `they inflate+docs` part). I imagine the solution could change and even be project independent, so maybe less of a fit for these more globally applicable guides?

davidstosik:
> Making a precaution in pull request, using commit status...?

The problem is that as soon as a file is commited to a branch and then pushed in a PR, it gets added to the repository, and cannot be removed easily. (Feel free to correct me if I'm wrong on this one.)

Checking PRs (manually or automatically) can help educating developers, but would happen too late to prevent the size increase to happen...

PR #52, opened 2017-08-08
Add Rake Tasks Section

halogenandtoast:
Add a section describing best practices for Rake tasks


Knack:
This is a good advice, but... what alternative would you suggest? For example, to put the methods inside a module or a class? 🤔

balvig:
@knack that's the idea I believe. Move as much logic as possible inside classes so that the rake task merely calls "run" on those. Might be nice to add an example to the guides?

Knack:
> Might be nice to add an example to the guides?

That's the point (and sorry for my rhetorical question) :wink:

balvig:
Be the change you want to see in the world 😉

PR #53, opened 2017-08-28
EnforcedStyle: template 

rikarumi:
#### Issue:
while working on https://github.com/cookpad/global-web/pull/5912 keep bumping to `%{something}` being check and mark as violated by hound.

#### This PR
~exclude spec from this delimiter checking to avoid a false positive issue~
EnforcedStyle: template to fix this issue

cc: @balvig


rikarumi:
updated the title and PR description.

Thank you @Knack!

PR #60, opened 2017-09-20
Rubocop: Warn about class/method length

Knack:
Reading the best practices, I noticed these recommendations about class and method lengths (#38):

- Prefer short, focused methods (aim for 1-liners, longer than 5 is a red flag)
- Prefer small, focused classes (100+ lines is a red flag)

To my surprise, those recommendations are disabled in our Rubocop configuration. I think we could give it a try and warn about these limits.

The default values in RuboCop are 100 lines for classes and 10 lines for methods. With these values, we currently have 10 warnings for classes and 61 for methods in `global-web`. I think it would be better to start with those default values (that's why I simply deleted those lines) because with the limit of 5 lines for methods our errors would rise to 414

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -49,9 +49,6 @@ CharacterLiteral:
 ClassAndModuleChildren:
   Enabled: false
 
-ClassLength:
-  Enabled: false
-
 ClassVars:
   Enabled: false
 
@@ -132,9 +129,6 @@ LambdaCall:
   Enabled: false
 
 LineEndConcatenation:
-  Enabled: false
-
-MethodLength:
   Enabled: false
 
 ModuleFunction:


davidstosik:
That seems like a good idea, but I wonder if there was any reason why it wasn't enforced in the first place, but just presented as a guideline/best practice... 🤔

Knack:
@davidstosik Well, I'm not sure of the origin of our RuboCop configuration but it was possibly copied from the configuration provided by Hound at that time.

And about why this type of cops was disabled by the folks at Hound, I have an assumption. Hound only alert for offenses in modified lines. I understand that it does this by running RuboCop on the whole file and then filters the errors, keeping only those that refer to a line of code that is among the modified ones.

The problem is that this type of offense does not refer to a particular line but to a group of lines. For example, if I add a new line in the middle of the body of a method, and that causes it to exceed the limit, the error shown by RuboCop may indicate the first line where the method is defined (the one containing the `def`) and not the line I've added, so Hound will ignore that message.

Actually, these types of "Metric" warnings will only be shown when in our commits we add new methods or classes that _from the beginning_ exceed the limit (or worse, if we modify the first line of a class or method that already exceeded it)

For this reason and others is why I do not like Hound too much 😒 . I mean, Hound is good, but I think that we should also run Rubocop in our development environment while we are writing code and **before** pushing our changes (well, I actually _do_  it 😅 )

Knack:
> we should also run Rubocop in our development environment while we are writing code and before pushing our changes

For example, here is a one-liner to run RuboCop on the files modified in your branch:

`git diff --name-only --diff-filter=d `git merge-base --fork-point master` | xargs -t rubocop`

balvig:
Heh, I'm actually a little against having this a part of the hound's setup 🙇  

Although I should say I'm not against trying it out for a while to see if I can be proven wrong, let me at least try to explain why I feel kind of hesitant about making this a hard rule:

- I think in order for the hound to be effective, and to avoid a "boy who cried wolf"-type scenario, I believe whatever the hound points out should as much as possible be "no-brainers", ie _just fix it_-type comments
- I therefore fear that if we allow for a lot of "I'm going to ignore the hound on this one" -type of situations, then slowly the importance of hound comments will be reduced and worst case turn into noise that people ignore
- For these particular rules, although I _do_ think they are a good guideline, I think they are a lot more case-by-base than some of the other syntax-related rules we have

A couple of examples would be some of the normalizers/sanitizers we have:

- [API remapper](https://github.com/cookpad/global-web/blob/86e893067f0ffdbb21952ebba2c42ab8cc099a9e/app/models/api/v1/remapper.rb#L7-L17)
- [Query normalizer](https://github.com/cookpad/global-web/blob/86e893067f0ffdbb21952ebba2c42ab8cc099a9e/app/models/query_normalizer.rb#L10-L17)

I think it's perfectly acceptable here to have the longer method (since it does a bunch of stuff) and I would hate for them to turn into something like

```ruby
def normalize
  convert_to_utf_8_and_uri_decode_and_downcase
  ...
  ...
end
```

in an attempt to work around the 5 line rule 😅 

The other alternatives I suppose are:

- Ignoring the hound's comments (and have them show up every time some one edits that code)
- Adding magical comments to disable that particular rule

I guess none of them feel like really satisfying solutions to me...? 🤔 

Perhaps I'm overlooking a way the examples above could be changed in a way that is an _improvement_, in order to comply with this rule?

Knack:
Agree 😅. As I said above, I just wanted to "give it a try" but, yes, it's adding too much noise.

I think this kind of metrics help you think about your code and how to refactor it better (for example, it has been useful to me to realize that I could improve some parts of the Newsletters code).

But as noted in the comments above, this type of checks does not fit well with Hound's way of reporting only the warnings that affect the modified lines. The same happens with other Metrics such as complexity, ABC size, Block length, etc.

Also, not all ruby file types must have the same rules (think of Rake tasks, migration files, specs, etc.)

I agree to revert this PR. :+1:

balvig:
@Knack alright! Will open the revert PR then to gather remaining opinions.

PR #61, opened 2017-09-27
Add a convention for using $stdin, $stdout, $stderr

JuanitoFatas:
Using the global `$stdout` over the constant `STDOUT`,
since other programs (like test suite) could change `$stdout`
to something else to make testing easier.

Also `STDOUT`, `STDERR`, `STDIN`'s default values are all to their `$std*` counterparts.

`$std*` docs: http://ruby-doc.org/core-2.4.2/doc/globals_rdoc.html


JuanitoFatas:
> how one would go to change what $stdout points to for the purpose of a test?

If you want to assert the output from programs that interact with command line.

PR #62, opened 2017-10-18
Revert "Rubocop: Warn about class/method length"

balvig:
Discussion [here](https://github.com/cookpad/guides/pull/60#issuecomment-337203807), cc @Knack 

@davidstosik @JuanitoFatas please pipe in if you have any additional comments/thoughts! 🙏

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -49,6 +49,9 @@ CharacterLiteral:
 ClassAndModuleChildren:
   Enabled: false
 
+ClassLength:
+  Enabled: false
+
 ClassVars:
   Enabled: false
 
@@ -129,6 +132,9 @@ LambdaCall:
   Enabled: false
 
 LineEndConcatenation:
+  Enabled: false
+
+MethodLength:
   Enabled: false
 
 ModuleFunction:


JuanitoFatas:
Instead of disabled for good, how about set a maximum of current status (`rubocop --auto-gen-config`)? 

>         --auto-gen-config            Generate a configuration file acting as a
>                                     TODO list.

So we can gradually improve.

Knack:
I'm using `.rubocop-todo.yml` in the Newsletters project. It is a good idea when you add Rubocop to an existing project, to gradually correct the offenses and focus on those new. However, I see several problems... 

* The first is that we would not only ignore the problems of "too many lines" but also ignore all other warnings (and we have _hundreds_ of them). 
* The second, you have to be committed to fix them someday. If not, this is simply a way to sweep the issues under the rug. 
* And finally, because of the way in which this file is generated, we can also hide new problems. An example: Suppose most of our offenses on method length have an average of 10-15 lines. However we have a special method with, say, 30 lines. When generating the TODO list we would be raising the maximum allowed to 30 lines, so if we create a new method of 20 lines we will not receive any warning.

Knack:
Having said that, I would agree to raise the maximum of 100/10 instead of simply disabling it. For example, with `--auto-gen-config` our maximums in global-web would be:

```
Metrics/ClassLength:
  Max: 228
Metrics/MethodLength:
  Max: 29
```

However, I think they are very high (the culprit is our `Recipe` model). We should put lower values. I personally would keep the limit of 100 for classes and maybe raise the limit on methods to 15.

balvig:
> Instead of disabled for good, how about set a maximum of current status
> So we can gradually improve.

@JuanitoFatas I guess my concern is that strictly adhering to "MAX 5 LINES PER METHOD" is not _always_ an improvement, and then we risk conditinoing ourselves to ignore the hound barking "sometimes" 😉 

In the end I feel like the entry in the guides is "good enough". Hopefully avoiding writing huge blocks of code should feel natural to developers even without the 🐶  to remind us!

PR #66, opened 2017-10-24
avoid trailing conditionals

balvig:
In general, the trailing conditional often feels like a bit of a "surprise" to me:

```ruby
UserMailer.new_photo_comment_reply(notification) if receiver_is_thread_owner?
```

It also increases line-length and kind of "hides" the conditional a bit, making the code _seem_ less "branchy" than it really is:

```ruby
if receiver_is_thread_owner?
  UserMailer.new_photo_comment_reply(notification)
end
```

One exception is perhaps guard clauses, where I think it makes sense to have them at the top for "aborting", separated by a blank line ie:

```ruby
def approve
  return if approved?
  return if unapprovable?

  update!(approved: true)
end
```

Thoughts?  I think this was originally suggested by the late halogenandtoast and I was reminded when seeing [this PR](https://github.com/cookpad/global-web/pull/6164/files#diff-54ae9546c5bd57393ed9bbc0d17a5caeR8) 😅  cc @kinopyo

### DIFF
--- a/best-practices/samples/ruby/4.rb
+++ b/best-practices/samples/ruby/4.rb
@@ -1,0 +1,19 @@
+## Bad
+def run
+  mark_related_items(:spam) if spam_detected?
+end
+
+## Good
+def run
+  if spam_detected?
+    mark_related_items(:spam)
+  end
+end
+
+## OK for guard clauses, separate by space
+def approve
+  return if approved?
+  return if unapprovable?
+
+  update!(approved: true)
+end
--- a/best-practices/samples/ruby/4.rb
+++ b/best-practices/samples/ruby/4.rb
@@ -11,5 +11,5 @@ def approve
   return if approved?
   return if unapprovable?
 
-  update!(approved: true)
+  update(approved: true)
 end
--- a/best-practices/samples/ruby/4.rb
+++ b/best-practices/samples/ruby/4.rb
@@ -1,13 +1,9 @@
 ## Bad
-def run
-  mark_related_items(:spam) if spam_detected?
-end
+mark_related_items(:spam) if spam_detected?
 
 ## Good
-def run
-  if spam_detected?
-    mark_related_items(:spam)
-  end
+if spam_detected?
+  mark_related_items(:spam)
 end
 
 ## OK for guard clauses, separate by space
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -6,6 +6,7 @@
 - Prefer small, focused classes (100+ lines is a red flag)
 - [Prefer extracting private methods over setting variables inside methods](samples/ruby/3.rb)
 - Prefer `$stdin`, `$stdout`, `$stderr` over `STDIN`, `STDOUT`, `STDERR`
+- [Avoid lines that end with conditionals (exception is guard clauses)](samples/ruby/4.rb)
 
 ## Rails
 


kinopyo:
Yeh I agree with the drawback of trailing condition and the exception in guard clauses 👍 

---

re @sebasoga's question, I also think it's a good idea to clarify the criteria for the guard clauses. I did a quick grep in our code base and dumped some examples as the base for discussions, see how we feel about it 🙂

```ruby
# a) "unfolded" condition
def show_halloween_logo?
  return false unless Region.current.path.in?(%w(fr it es))
  ...
end

# a') "unfolded" condition - long version
def comment_groups
  return [] unless object.target.class.in? [Comment, PhotoComment]
  ...
end

# b) "folded" condition
def show_halloween_logo?
  return false unless in_halloween_season?
  ...
end

# b') "unfolded" condition - long version
def comment_groups
  return [] unless commentable?
  ...
end

# c) slightly long return value, but is one "unit" of mental block that is easy to parse
def detailed_premium_status(user)
  return t("admin.premium_services.status.free") unless user.premium?
  ...
end

# d) long return value, but is requires mental energy to parse
def search_title(query:, count:)
  return t("searches.show.title_tag_no_results", query: query).mb_chars.capitalize.to_s if count == 0
  ...
end
```

balvig:
@sebasoga I think it almost depends on what `foo?` and `bar?` etc is 😅  

For example, the exception to the rule made for guard clauses is _not_ meant to encourage something like:

```ruby
def status
  if admin?
    :admin
  else
    :user
  end
end
```

...being refactored into

```ruby
def status
  return :admin if admin?
  :user
end
```

I guess guard clauses are more about "aborting" early on a certain condition, as the result of that would make rest of the method void, ie:

```ruby
def upload_image
  return if image.blank?
  return if file_server_down?

  copy_image_to_temporary_folder
  copy_temporary_folder_to_file_server
end

def account_balance
  return 0 if bank_accounts.none?

  bank_accounts.do_some_crazy_calcumalating
end
```

etc...

So looking at the code @kinopyo escavated, I think there _may_ be a chance that some of the examples above don't fit that description (I think I would have to go in and look at the snipped `...` parts to really get a feel though 😅  🙇 )

sebasoga:
@balvig probably just adding the examples you just described above as examples of what not to do, would be good enough to cover a lot of those cases.

best-practices/samples/ruby/4.rb
```diff
@@ -0,0 +1,19 @@
...
+def run
+  mark_related_items(:spam) if spam_detected?
+end
+
+## Good
+def run
+  if spam_detected?
+    mark_related_items(:spam)
+  end
+end
```
Knack: BTW: I'm not sure if this is a good example, I'd probably rewrite it as:

```ruby
 def run
   return unless spam_detected?

   mark_related_items(:spam)
 end
```

Why don't you just leave the affected code?

```ruby
## Bad
  mark_related_items(:spam) if spam_detected?

## Good
  if spam_detected?
    mark_related_items(:spam)
  end
```

Knack:
__my 2¢__: I have to admit that I like trailing conditionals, although I understand the difficulty to read it with long or complex expressions. For me it would be better to say something like:

"Avoid lines that end with conditionals _except in cases where all expressions are extremely trivial_"

Things like:

```ruby
yield if block_given?

log_fully_accepted_chat if all_members_accepted?

redirect_to root_path if logged_in?

destroy if cooking_photos.none?
```

Or even :smile:
```ruby
mark_related_items(:spam) if spam_detected?
````

File: best-practices/samples/ruby/4.rb [hunk 2]
```diff
@@ -0,0 +1,15 @@
...
+if spam_detected?
+  mark_related_items(:spam)
+end
+
+## OK for guard clauses, separate by space
+def approve
+  return if approved?
+  return if unapprovable?
+
+  update!(approved: true)
```
JuanitoFatas: Should be `update(approved: true)`? since the method defined here doesn't have bang suffix.

balvig: Heh, I can remove to reduce confusion, but that's actually not really the convention we follow for bang methods. http://www.chrisrolle.com/en/blog/dangerous-prima-donna-methods

balvig:
> Avoid lines that end with conditionals except in cases where all expressions are extremely trivial

@Knack I see. 🤔  What do you think about starting out with the simpler/stricter rule, then in the future whenever we see a place where we see it would help readability to diverge from the rule, highlighting it in the PR so we can revisit?

PR #67, opened 2017-11-01
Enforce request.referrer over request.referer

Knack:
## referer or referrer?

[One of the offenses](http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/RequestReferer) detected by Rubocop is the consistency in the use of `request.referer` or` request.referrer` throughout the code. Rack [defines both methods](https://github.com/rack/rack/blob/b0818268fd9f4026a1b0fb205cc1060ee38a23cc/lib/rack/request.rb#L141-L142) with an alias. To know more about why the misspelled word `referer`is used, you can consult the [wikipedia](https://en.wikipedia.org/wiki/HTTP_referer#Etymology)

Checking our `global-web` project, the use of each word is quite divided:

    Referrer: 7 usages in 4 files
    Referer:  7 usages in 6 files

I have created this PR as a place to discuss our preferences. It is a pull request instead of an issue because I personally would prefer to change the default value and use `referrer` (I am a _grammar nazi_), although I understand the opposite position to maintain consistency with the HTTP specification

_What is your opinion?_

(Of course, another option would be to ignore all this and disable the Cop)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -309,3 +309,6 @@ Rails/Output:
 
 Rails/SkipsModelValidations:
   Enabled: false
+
+Rails/RequestReferer:
+  EnforcedStyle: referrer


.rubocop.yml
```diff
@@ -309,3 +309,6 @@ Rails/Output:
 
 Rails/SkipsModelValidations:
   Enabled: false
+
+Rails/RequestReferer:
+  EnforcedStyle: referrer
```
JuanitoFatas: +1 on referer. 1 less character to type and:

> Make appropriate exceptions for industry standard nomenclature/jargon, including plain misspellings. For instance:
> Use "referer", not "referrer", in the context of the HTTP protocol.
> https://google.github.io/styleguide/lispguide.xml?showone=Spelling_and_Abbreviations#Spelling_and_Abbreviations

balvig:
I'm okay with either one, but I think good idea to be consistent 👍 
Am personally okay deferring to whoever feels the strongest about which one to use 🙏

kinopyo:
I'm with @Knack, I think. Until now I haven't thought it that much, but whenever I type it, my mind always seem to transfer it to the double R `referrer` to my fingers 😅 

And since the shorter one `referer` is indeed a misspell of the original true English word "referrer", I don't know if I'd like to indulge it anymore 😅 

---

re character size) One interesting thing I found about is, maybe the pure character length is not _always_ a good measure (in terms of typing). Remember the `have_text` vs `have_content` issue?
 One of its sales pitch was less characters, it looks good on paper, but when I tried to type it on the keyboard, I noticed the word "have_text" relies heavily on the left hand and the travel distant is far (a underscore follows the word "text"), reminds me of swiping 200 units of armies in Starcraft 😅 

In that sense a single typing a single "r" or a double "rr" itself is not a concern for me. 🙂

(Just my feeling towards the word "referrer" and the experience of typing it on the keyboard)

Knack:
Any other opinion on this?

PR #70, opened 2017-12-15
Allow all our Rails environments in Rubocop

davidstosik:
[Rubocop defaults.](https://github.com/bbatsov/rubocop/blob/c3ceec9d982fa27d51c5ff8a159285b910b85013/config/default.yml#L1660-L1664)

[Example of Hound failure.](https://github.com/cookpad/global-web/pull/6555#discussion_r157256452)

image


davidstosik:
Environments are project-dependant, moving to global-web: #6556

PR #80, opened 2018-04-23
[WIP] disable Rails/DynamicFindBy cop

balvig:


davidstosik:
I thought the same recently, then I realized that the cop's value is in suggesting me not to use a name that could induce the next developer reading that section of code.

As I read `IdentityProviderGateway.find_by_full_name(capitalized_provider_name)`, my first expectation is that `IdentityProviderGateway` has a `full_name` attribute.

If it is an ActiveRecord model, I'm gonna expect a `full_name` column. If it's not an AR model, I guess this is less relevant, but still, I think that renaming `find_by_full_name` for something that does not match some convention would make the code clearer?

`IdentityProviderGateway.for_full_name`, `IdentityProviderGateway.with_full_name`, `IdentityProviderGateway.first_with_full_name`, etc...

balvig:
@davidstosik sorry that I used that example! That is a refactoring in progress, and there are problems with the existing implementation that make it a bit more awkward to work with 😅 

I guess this proposal isn't so much to discuss the specific example, but more in general the problem with this cop (I better example would have been `ActiveHash` models!)

balvig:
@davidstosik @Knack 

While I see the point, I somehow find it a little hard to agree 😅  🙇 

Ie for one I think it would be quite tricky to draw a line as to what methods are allowed to share the same names as ActiveRecord? 🤔 

How about these for example?

```
.find
#save
#update
#create
#destroy
```

Or how about models like Spyke, where the goal is to approximate AR, but it still doesn't have the full set of methods? I feel like these _and_ `find_by` methods should be available to us as vocabulary whether we're designing an AR-backed class or a PORO?

I guess also I find it hard to think of a situation where it would be a problem for me whether I know it's an AR method or not, but maybe I'm overlooking something... 🤔  

To me the data store is in a way an implementation detail: What I care about is the public API and whether that makes sense!

Ie imagine transitioning from `ActiveHash` to `ActiveRecord` for example. 

I believe that should preserve most of the public API and we shouldn't have to rename `#by` methods to `#find_by` just because we've switched the data store from YML to DB?

Knack:
Anyway, I checked some of our projects and currently we don't have any offense of this kind. So maybe it's just _you_ who is determined to use that name? 😸

balvig:
@Knack heh, perhaps people avoid them because the dog barks at them and we end up with [compromises](https://github.com/cookpad/global-web/blob/71db80f2c83ce998ae0aec01ff94a97d56da886c/app/models/user.rb#L107-L113) 🐶 😉 

I currently have a specific use case where I want to use them in https://github.com/cookpad/global-web/pull/7561. I guess I don't see much benefit to changing the finder to `.by`, in fact I even think it would make it slightly more vague 🤔 

In any case I believe the intention of the rule is simply preferring `find_by(name:)` over `find_by_name` _for ActiveRecord models?_

I think what I'm hearing above is a proposal for a _different_ rule that says something like "don't use ActiveRecord-like methods in non-ActiveRecord objects"?

So while as mentioned I find that a little hard to agree with (I find it's often even _desirable_ to approximate ActiveRecord methods to get a consistent interface! 😅) if you want to open a _separate issue_ making a case, we can certainly have that discussion.

For this PR however, considering that we don't have any established practice that says "don't user AR-like methods", I still believe that this cop is basically a distraction.

balvig:
Thanks @Knack, merging this! @davidstosik let me know if you want to have that other discussion sometime 🙏

PR #81, opened 2018-05-01
Avoid logic in rake tasks

lewispb:
This follows feedback on this PR:

https://github.com/cookpad/global-web/pull/7615/files/3aa28f6d5d4709b5469b0ac15afeb5c20b490d48#diff-f3c37f84a144278a1a9cdbeed263677d

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -407,10 +407,32 @@
 
 ## Rake Tasks
 
-- <a name="avoid-methods-inside-rake-namespace"></a>
-  Avoid defining methods inside of Rake namespaces
-  This will define methods at the top level and interfere with things like factories
-  <sup>[link](#avoid-methods-inside-rake-namespace)</sup>
+<a name="avoid-methods-inside-rake-namespace"></a>
+- Avoid defining methods inside of Rake namespaces. This will define methods at the top level and interfere with things like factories.
+- Avoid writing logic in rake tasks, use a class instead.
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    namespace :users do
+      desc "Decrements all users rank"
+      task decrement_rank: :environment do
+        User.where("rank > 0").update_all("rank = rank - 1")
+      end
+    end
+
+    ## Good
+    namespace :users do
+      desc "Decrements all users rank"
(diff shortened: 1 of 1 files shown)


best-practices/README.md
```diff
@@ -407,10 +407,32 @@
...
+      task decrement_rank: :environment do
+        User.where("rank > 0").update_all("rank = rank - 1")
+      end
+    end
+
+    ## Good
+    namespace :users do
+      desc "Decrements all users rank"
+      task decrement_rank: :environment do
+        User.all.decrement_rank!
```
davidstosik: [Nitpick] Should we make sure that "good" examples would pass a code review?

If I were to review this code in a PR, my comment would probably go something like this:

> It seems you probably want a `User.decrement_all_ranks!` kind of method? What does calling it on `.all` introduce?

lewispb: Thanks for the feedback especially as I am actually going to submit this as a PR to global-web 😄

best-practices/README.md
```diff
@@ -407,10 +407,32 @@
...
+- Avoid writing permanent logic in rake tasks, use a class instead.
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    namespace :users do
+      desc "Decrements all users rank"
+      task decrement_rank: :environment do
+        User.where("rank > 0").update_all("rank = rank - 1")
```
karlentwistle: I am a bit confused because the description says "Avoid defining methods inside of Rake namespaces" but I can't see a single method defined inside the bad example. Since `where` and `update_all` are both already accessible on `User`. 🤔

lewispb: That was the existing guide. I added a new guide and an example to go with it.

karlentwistle: Ahhhhhhhhhhhhh 💡

best-practices/README.md
davidstosik: Good call on removing the `!` too! 👍

best-practices/README.md
sebasoga: @lewispb, I think @karlentwistle has a point, the link is now placed in a weird place after the wrong bullet point.

PR #83, opened 2018-05-03
Use minutes not second when asserting results order

sebasoga:
This is a follow-up to a discussion started in https://github.com/cookpad/global-web/pull/7716#pullrequestreview-116719537

In order to avoid flaky tests that fail based on the second the test gets executed as explained by @karlentwistle in the image below:

We should consistently use minutes, another option as mentioned by @knack before would be to freeze time.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -829,6 +829,30 @@
     expect(page).to have_text t("signups.create.welcome_message")
     ```
   </details>
+  
+  - <a name="use-minutes"></a>
+  Use minutes instead of second when a setting timestamps to asserting on results order
+  <sup>[link](#use-minutes)</sup>
+
+  <details>
+    <summary><em>Example</em></summary>
+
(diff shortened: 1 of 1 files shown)


best-practices/README.md
```diff
@@ -829,6 +829,30 @@
     expect(page).to have_text t("signups.create.welcome_message")
     ```
   </details>
+  
+- <a name="use-minutes"></a>
+  Use minutes instead of second when a setting timestamps to asserting on results order
```
karlentwistle: This might read slightly better like this:

Use minutes instead of seconds when setting timestamps for the purpose of making assertions on ordering of results.

lewispb: ... the purpose of asserting the order of results.

😄

sebasoga:
@karlentwistle, @lewispb thanks for your suggestions.

PR #84, opened 2018-05-10
Prefer 3.times.map over Array.new(3).map

lewispb:
Context https://github.com/cookpad/global-web/pull/7808#issuecomment-387919152

Disable https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Performance/TimesMap

I prefer the readability of `3.times.map { |x| }` over `Array.new(3).map { |x| }`.

More discussion: 
https://stackoverflow.com/questions/41518896/why-does-rubocop-suggest-replacing-times-map-with-array-new

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -270,6 +270,9 @@ Lint/ParenthesesAsGroupedExpression:
 Performance/RedundantMerge:
   Enabled: false
 
+Performance/TimesMap:
+  Enabled: false
+
 Lint/RequireParentheses:
   Enabled: false
 


lewispb:
@Knack sorry my mistake on the syntax.

Knack:
>  I found the original diff was easier to follow that way.

But that's because the original code had a **bug**

`Array.new(3).map {}` means "Create an empty array and use it to map _another_ array"

Whereas

`Array.new(3) {}` means "Create an array of size 3 and initialize it". Much simpler, IMHO.

Even if all the values are the same you can pass it as a second argument: `Array.new(3, "x")`

But as I said, I don't see a big difference and I don't care too much about performance. So it's ok to disable it.

balvig:
Thanks for the extra explanation @Knack!

I also agree with this quote:

> Either one seems good

In the end I think the PR title is maybe incorrect, but the fix is 👍 :wink:

PR #88, opened 2018-06-01
Add a guide for service objects in Rails

lewispb:
Add a guide for service objects in Rails

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -296,3 +296,58 @@
 - <a name="dont-use-console"></a>
   When _reading data_ from production, avoid using the console, use redash instead.
   <sup>[link](#dont-use-console)</sup>
+
+- <a name="service-object-method"></a>
+  Service objects should have a single public method `#run`, which does not accept arguments
+  <sup>[link](#service-object-method)</sup>
+
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    class ChatMessage
+      def run(chat:, :body)
+        chat.messages.create(body: body)
+      end
+    end
+
+    ## Good
+    class ChatMessage
+      def initialize(chat:, body:)
+        @chat = chat
+        @body = body
+      end
+
+      def run
+        chat.messages.create(body: body)
+      end
+
+      private
+        attr_reader :chat, :body
+    end
+    ```
+  </details>
+
+- <a name="service-object-return"></a>
+  If a service object needs to return a value, prefer a meaningful object, e.g. an invalid model over false
+  <sup>[link](#service-object-return)</sup>
+
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    def run
+      message = chat.messages.new(body: body)
+      message.save # true/false
+    end
+
+    ## Good
+    def run
+      message = chat.messages.new(body: body)
+      message.save
+      message
+    end
+    ```
+  </details>


karlentwistle:
This LGTM @lewispb thanks very much for writing this. Do you think it might be worth mentioning the `run` method when using your service.

For example:

Good:

```ruby
AwesomeService.run(foo: "bar")
```

Bad:

```ruby
AwesomeService.new(foo: "bar").run
```

best-practices/README.md
```diff
@@ -405,6 +405,52 @@
     ```
   </details>
 
+- <a name="service-objects"></a>
+  To avoid complex controller actions or model callbacks, consider using a service object. <sup>[link](#service-objects)</sup>
+    - Accept arguments in the initializer
+    - Inherit from `ApplicationService`
+    - Have a single public method `#run`, which does not accept arguments
+    - Inject all dependencies
```
balvig: I'm a little unsure of what this means in practice? For example I think something like

```rb
def logger
  ActivityLogger.logger
end
```

might be ok, but not sure if the suggestion is to

```rb
def initialize(...., logger: ActivityLogger.logger)
```

which I _could_ imagine being a bit more case by base?

best-practices/README.md
```diff
@@ -405,6 +405,52 @@
...
+
+    ## Good
+    class ChatRenamer < ApplicationService
+      def initialize(chat:, new_name:)
+        @chat = chat
+        @new_name = new_name
+      end
+
+      def run
+        chat.update(new_name: new_name)
```
balvig: Wonder if this goes against the advice above?

"Return a meaningful object, e.g. prefer an invalid model over `false`"

best-practices/README.md
```diff
@@ -405,6 +405,52 @@
     ```
   </details>
 
+- <a name="service-objects"></a>
+  To avoid complex controller actions or model callbacks, consider using a service object. <sup>[link](#service-objects)</sup>
```
balvig: I feel like maybe there is a _tiny_ gap between what this first line says, and what the rule is actually about.

Ie I think I would find myself linking to this when someone was _already_ using a service object, but not following our conventions (I think that's how this came about?)

Maybe this is actually two new rules?

- Service objects should have a single public `run` method, which does not accept arguments
- Service objects should return something meaningful

Each could have their own code example?

best-practices/README.md
lewispb: Yeah, let's lose that suggestion.

best-practices/README.md
lewispb: I thought this introduction made sense because we don't have anything regarding service objects in the guide yet.

> I think I would find myself linking to this when someone was already using a service object, but not following our conventions

What about when someone had written a complex controller action or duplicated logic in an admin controller and public facing controller?

We could add multiple rules but I think the way I've done it here I feel covers several scenarios and is fairly succinct. Up to you though?

balvig: > What about when someone had written a complex controller action or duplicated logic in an admin controller and public facing controller?

Ah yes, I feel like that is actually a _different_ "linkable" rule we may/may not need, and it might even be slightly ambiguous/controversial, as I think that depending on the case there might be _other_ solutions than service objects (ie extract to method etc). 

My hope is that most of these we can actually let a CI do, so granularity is good!

Considering the actual problem that inspired this was "unorthodox" usage of a service object (object had readers on it), I would prefer to be able to add a comment on the actual line with the problem, pointing to the actual rule that tells the submitter exactly what's wrong.

Hope that makes sense!

best-practices/README.md
```diff
@@ -405,6 +405,52 @@
...
+    - Accept arguments in the initializer
+    - Inherit from `ApplicationService`
+    - Have a single public method `#run`, which does not accept arguments
+    - Return a meaningful object, e.g. prefer an invalid model over `false`
```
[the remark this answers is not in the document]
lewispb: I was inspired by this post:

https://hackernoon.com/the-3-tenets-of-service-objects-c936b891b3c2

> Think about it this way. There are more than just two outcomes for a service object. 1. Success. 2. Failure. 3. Unhandled Exception(s). It would be a disservice to your service to handle such outcomes just in a boolean response.

> When taking these outcomes more in depth, there is, even more, a reason for a response object. Let’s take the success example. Firstly, a result object that can hold a success? Predicate method to handle conditional branching for controller responses, and more. Additionally, since it’s your own response object you could name the success/failure method how you would like. If signed_up? makes more sense, it’s as simple as defining it in your object.

> So, how can you use a response object? The easiest use can will normally be using a Struct or OpenStruct from the Ruby library.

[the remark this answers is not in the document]
Knack: What I mean is that the phrase seems to imply that you **must** return something, and if so, I don't agree. Most of our services (those of the type _"things that happen after ..."_) do not return any value and I find it totally valid, even recommended.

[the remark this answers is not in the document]
lewispb: Thanks @Knack , could you post examples?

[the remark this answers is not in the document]
lewispb: Maybe there's an argument to be made that a service like that should check the result of everything it runs and aggregates the results? Some may be optional, some may not be. Perhaps another caller in the future will need to know if the service successfully performed its duties or not?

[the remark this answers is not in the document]
Knack: > Perhaps another caller in the future...

[YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)? 😅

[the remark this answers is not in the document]
balvig: I think I agree with @Knack that I don't see anything in our _current_ codebase, that would push us to say "_all_ services must return _something_". 

I also kinda like how not returning anything works well with the event sourcing concept. Services without return values will be a lot easier to extract if needed?

In any case, I don't see any harm (yet!) in softening the phrasing a bit?

 Maybe something like:

> If `run` needs to return a value, prefer a meaningful object, e.g. an invalid model over `false`

What do you think @lewispb?

best-practices/README.md
```diff
@@ -451,6 +444,50 @@
     ```
   </details>
 
+  <details>
```
Knack: Should we add a description to this second example? Something like:

"Service objects should be executed by calling its class method `run`, defined in the base class"

balvig:
@lewispb since I recently reorganized the repo #92, let me try wrapping this up

balvig:
@lewispb in the end, I ended up paring this down to the two rules, hope it makes sense.

I felt a little uncomfortable about adding the others as:

- `ApplicationService`: Only exits in `global-web` atm and its current purpose is to provide convenience to "Call class method vs instantiate". In the end I don't feel like it's a big of enough deal whether we do one or the other as long as we follow the rules above, which prevents us from f.ex instantiating a service in order to call accessors/other methods than `run`.
- `Consider using a service object`: There is something about this that makes it feel a little awkward to me. I think maybe it's the fact that it's kind of suggesting a pattern, but a lot less clear cut than f.ex. "Avoid using ternary operator", and I could imagine someone linking to the rule but without it having much to offer besides "consider it". I guess I think more of it as "avoid complex controller actions" or "avoid repeated logic", with a service object just being _one_ approach (among many).

Let me know if you want to discuss them a bit more in realtime! 🙇

PR #89, opened 2018-06-05
Avoid HTML in locale file

firewalker06:
@balvig referring to our discussion in https://github.com/cookpad/global-web/pull/8143#discussion_r192950014

This suggests to avoid including HTML tags in translations. HTML tags increases weariness of translators, who is often unfamiliar with it.

Even though the example is coming from Admin land where there is no active translations going on, this suggestion would be good in general projects.

### DIFF
--- a/best-practices/README.md
+++ b/best-practices/README.md
@@ -868,9 +868,9 @@
   Don't use `zero:` key to display a "no results" message (it is intended only to allow proper grammar)
   <sup>[link](#dont-abuse-zero-key)</sup>
 
-- <a name="avoid-html-in-locale"></a>
-  Avoid including HTML in locale file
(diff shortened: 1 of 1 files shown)


best-practices/README.md
```diff
@@ -868,6 +868,36 @@
   Don't use `zero:` key to display a "no results" message (it is intended only to allow proper grammar)
   <sup>[link](#dont-abuse-zero-key)</sup>
 
+- <a name="avoid-html-in-locale"></a>
+  Avoid including HTML in locale file
```
balvig: I think we can be even stronger in the language here, ie "Don't include HTML in locale files" 😉

firewalker06:
Good point, maybe the HTML tag we use now will be deprecated in HTML6 😅.

PR #92, opened 2018-06-11
Reorganize to focus just on styles

balvig:
Closes https://github.com/cookpad/guides/issues/91

This: 

- Splits up best practices into different files
- Merges I18n/database operations into "Rails"
- Removes "workflow" and "code review", I'm preparing a new place to host these "non-style" related topics!

Note, am planning to:

- Open a "web chapter" repo to capture some of the other practices/values we have in the web chapter we have outside of code
- Rename this repo to `global-style-guides` to try and reduce some of the confusion that happens with the existance of https://github.com/cookpad/styleguide 😅 
- Invite @cookpad/global-ios and @cookpad/global-android members to start keeping their style guides here  (as far as I understand Android doesn't use a style guide though?)


davidstosik:
> Python

I guess they already have [PEP-8](https://www.python.org/dev/peps/pep-0008/)! 😄

balvig:
> PEP-8

Is that @cp-8’s cousin?

PR #95, opened 2018-06-18
Add conventions about testing endpoints

sebasoga:
I am suggesting the following:

- Let's keep using `example` everywhere for request specs. Splitting between `it` and `example` only made sense when tests were mixed in the same file. Now that documenting tests go in a separate file there's no point in using a different part of the DSL which [have the same effect](https://github.com/zipmark/rspec_api_documentation#example).
- Let's stop using `document: true`. AFAIK the `document` meta tag is not really being used to pick tests will be included in the documentation and which not. Moving forward we will choose the test that should generate documentation based on the path they're in so there's even less reason to use the meta tag.
- Let's make it a requirement to not only start adding _documenting tests_ for new endpoints but also to document their parameters and response. The point of creating documentation is for it to be complete, not just a list of available endpoints that client applications need to start guessing how to use them

@cookpad/global-search what do you think about starting to document you API? AFAIK there's no documentation for the API we use to connect `global-web` to `global-search` every time I need to integrate with `global-search` I need to talk to someone working on that application to check if the endpoint I need already exists and how to use it. I'd be happy to help get you started.

@davidstosik what do you think about doing the same for `global-payments`?

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -371,13 +371,17 @@
   </details>
   
 - <a name="testing-endpoints"></a>
-  Do not use the `document` meta tag when writing tests for endpoints
+  Do not use the `document` meta tag when writing tests for endpoints. Tests in the
+  `specs/requests/api/` path are intended to create documentation for endpoints but to 
+  test the assert on the endpoint response or side-effects.
   <sup>[link](#testing-endpoints)</sup>
 
   <details>
     <summary><em>Example</em></summary>
 
     ```ruby
+    ## spec/requests/api/bookmarks_spec.rb
+    
     ## Bad
     resource "Bookmarks" do
       get "/:version/users/:user_id/bookmarks" do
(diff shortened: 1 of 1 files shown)


balvig:
Hey @sebasoga, good job starting on this!

I had a vague idea that maybe we could start with `README.md` files inside each spec directory in the app itself, since as you point out this is mostly a ["try-out"](https://github.com/cookpad/web-chapter/issues/10) in `global-web` at the moment, and it even relies on [some custom code from that app](https://github.com/cookpad/global-web/commit/9102dfa2c3dd58ba8f4162e621cf9572ef87a95e) to work.

I do realize that we risk a little bit of "where are the docs?"-effect, but the thinking I had was it could start there and if we find it to be a good pattern we want across all apps we could extract to gems/global style guides as needed.

I'd be happy to work on this with you since also I keep asking questions about it 😅 

Not sure what your whereabouts are at the moment but perhaps we can find an overlapping time sometime this or next week?

sebasoga:
@balvig I'm back home. This week will be busy for me, so I'd be glad to schedule some time to work on this next week. Any day 9am JPT (7pm local) would work for me.

balvig:
(invite sent, thanks!)

sebasoga:
I believe this can be closed after working in https://github.com/cookpad/global-web/pull/8447.

PR #96, opened 2018-06-25
Document some usages of CONSTANTS

balvig:
Pretty basic, but [since it came up](https://github.com/cookpad/global-web/pull/7394#discussion_r179053151), this documents a couple of scenarios where using CONSTANTS is encouraged.

cc @tapster

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -156,3 +156,63 @@
     ```
   </details>
 
+- <a name="explanatory-constants"></a>
+  Describe unexplained strings/numbers using constants
+  <sup>[link](#explanatory-constants)</sup>
+
+  <details>
+    <summary>Example</summary>
+
+    ```ruby
+    ## Bad
+    def params
+      { gak: "UA-4235" }
+    end
+
+    ## Good
+    def params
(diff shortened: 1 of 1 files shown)


sebasoga:
I've had mixed feelings about things like:
```rb
def google_analytics_key
  "UA-4235"
end
```

Is this something we want to encourage or discourage?

balvig:
@sebasoga  What are the mixed feelings about?

sebasoga:
Is this?
```rb
KEY = "UA-4235".freeze

def params
  { google_analytics_key: KEY }
end
```
better than
```rb
def params
  { google_analytics_key: google_analytics_key }
end

def google_analytics_key
  "UA-4235".freeze
end
```

balvig:
@sebasoga Ah I see. I guess it depends on whether the method might be overridden in child classes or quacking with a mix of dynamic/static responses to help out in views, that kind of thing?

```rb
class Recipe
  def summary
    “Recipe about #{title}”
  end
end

class BlogPost
  def summary
    “This is just a blog post”
  end
end
```

PR #99, opened 2018-07-05
Add testing guide about writing tiny specs

sebasoga:
Worked in this with @balvig during our pairing session earlier this week. We wanted to add this convention which we extracted from this PR: https://github.com/cookpad/global-web/pull/6019.

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -370,7 +370,7 @@
     ```
   </details>
   
-  - <a name="use-minutes"></a>
+- <a name="use-minutes"></a>
   Prefer splitting up longer specs into tinier ones, even if they require the same setup
   <sup>[link](#use-minutes)</sup>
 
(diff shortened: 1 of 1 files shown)


testing/README.md
```diff
@@ -369,3 +369,49 @@
     expect(Recipe.recently_published.last).to eq(older_recipe)
     ```
   </details>
+  
+- <a name="use-minutes"></a>
```
balvig: "Minutes"? (maybe a usage of that word I don't know)

testing/README.md
```diff
@@ -369,3 +369,49 @@
...
+  Prefer splitting up longer specs into tinier ones, even if they require the same setup
```
balvig: Nitty suggestion:

```
Prefer splitting up longer specs and give each scenario its own spec, even if they require similar setup.
```

testing/README.md
sebasoga: This was a copy-paste problem, fixed!

PR #103, opened 2018-08-02
Add Redis guides

sebasoga:
Thought it would be a good idea to document some guidelines we should all keep in mind.

### DIFF
--- a/redis/README.md
+++ b/redis/README.md
@@ -1,0 +1,8 @@
+# Redis
+
+- <a name="dont-use-keys-in-production"></a>
+  Redis is threaded this means some commands can potentially block the instance while being 
+  executed, such is the case for `KEYS` (`redis.keys(*)`), look for alternative operation 
+  like `SCAN`. When in doubt look at [Redis' oficial docs](https://redis.io/commands) or 
+  consult with the [SRE team](https://github.com/orgs/cookpad/teams/infra/members).
+  <sup>[link](#dont-use-keys-in-production)</sup>


cintamani:
Shall we open a RFC issue and start talking about how we want to use Redis? 

For example, if Redis is used as a cache, we want to make sure that: 
 - All keys are expired and renewed at request 
 - Redis can go down without impacting user experience 

If redis is used as a store, we want to make sure: 
 - We analyse in advance the memory usage of the instance

sebasoga:
@cintamani that sounds like a good idea! A way of doing that could be by creating another adding more guidelines to the `redis/README.md` file this PR introduces.

redis/README.md
```diff
@@ -0,0 +1,8 @@
+# Redis
+
+- <a name="dont-use-keys-in-production"></a>
+  Redis is threaded this means some commands can potentially block the instance while being
```
errm: single threaded

redis/README.md
```diff
@@ -0,0 +1,8 @@
+# Redis
+
+- <a name="dont-use-keys-in-production"></a>
+  Redis is threaded this means some commands can potentially block the instance while being 
+  executed, such is the case for `KEYS` (`redis.keys(*)`), look for alternative operation 
+  like `SCAN`. When in doubt look at [Redis' oficial docs](https://redis.io/commands) or
```
l15n: [nit: spelling] `official`

redis/README.md
balvig: Can we shorten this to a one-line, a la the other guides? Ie basically "Don't use `keys` in production". Then perhaps it's enough to keep the "why" in the PR?

redis/README.md
cintamani: I would mention tha `scan` is a very slow command ( O(N) with N being the number of keys in the instance). We don't want any code to rely on a scan, just one-off or background job tasks

sebasoga:
Please ♻️

Issue #90, opened 2018-06-06
Add guide on bang / non-bang methods


lewispb:
http://davidablack.net/dablog.html#2007/8/15/bang-methods-or-danger-will-rubyist

davidstosik:
Yes! 👍 
No [_Prima Donna_ methods](http://www.chrisrolle.com/en/blog/dangerous-prima-donna-methods), please!

I like this quote which sums it up well:
> _Any method can be destructive._ The `!` is solely to convey information for **dangerous versions of existing methods**. ([source](https://schwad.github.io/ruby/rails/code-quality/code-smells/2018/02/14/coming-to-terms-with-the-prima-donna-method-smell.html))

cp-8:
[BEEP BOOP] Hi there!

This issue/PR hasn't been updated for _4 weeks_ so closing for now.

Feel free to re-open in the future if/when it becomes relevant again! :heart:

---

_Configure by adding `?config[stale_issue_weeks]=...` to the webhook._

sikachu:
Come back!!

balvig:
@sikachu maybe you could have a go at a PR? 🙏

davidstosik:
(@giga811 FYI, as we talked about it today. 😉 )

cp-8:
[BEEP BOOP] Hi there!

This issue/PR hasn't been updated for _4 weeks_ so closing for now.

Feel free to re-open in the future if/when it becomes relevant again! :heart:

---

_Configure by adding `?config[stale_issue_weeks]=...` to the webhook._

cp-8:
[BEEP BOOP] Hi there!

This issue/PR hasn't been updated for _4 weeks_ so closing for now.

Feel free to re-open in the future if/when it becomes relevant again! :heart:

---

_Configure by adding `?config[stale_issue_weeks]=...` to the webhook._

PR #104, opened 2018-08-03
Add guide regarding use of bang methods

aadityataparia:
closes https://github.com/cookpad/global-style-guides/issues/90


ruby/README.md
```diff
@@ -60,6 +60,35 @@
...
+      # destructive
+    end
+
+    ## Good
+    def method
+      # destructive
+    end
+
+    ## Good
+    def method!
```
sikachu: How do people usually sort these methods btw? I feel like non-bang should come before bang?

aadityataparia: In http://davidablack.net/dablog.html#2007/8/15/bang-methods-or-danger-will-rubyist, it bang is before non bang, I guess it depends on preference unless we already have a general method sorting guideline.

Personally I like dependent method above methods that are used in it. Which feels ruby like - public method are above private methods :P

l15n:
> Avoid using bang methods i.e. ! at the end of method names, unless you want to convey that bang method is dangerous version of an existing method.

This is known by some as the "[prima-donna method smell](https://schwad.github.io/ruby/rails/code-quality/code-smells/2018/02/14/coming-to-terms-with-the-prima-donna-method-smell.html)"... but I think the key point is _not to avoid using bang methods_, rather avoid _naming_ methods with a _bang_ without a corresponding non-bang method.

To capture the above subtlety, perhaps you could consider rewriting the rule to emphasise that this is about _naming_ and _defining_ methods, rather than _using_ methods.

e.g.

> Avoid method names ending in `!` ("bang") without a corresponding bang-less, safe version of the method.

ruby/README.md
l15n: The above-mentioned link doesn't really say that the bang method should go _before_ the non-bang method. Rather, 

>  It’s conventional to define the non-bang method in terms of the bang one. 

This convention implies that the bang-method goes first _only_ if our convention for _ordering_ method definitions was dependencies-first.

I _think_ we currently prefer to write _dependents_ first, i.e you scroll down to find dependencies of the current method.

ruby/README.md
```diff
@@ -60,6 +60,35 @@
...
+  Avoid using bang methods i.e. `!` at the end of method names, unless you want to convey that bang method is dangerous version of an existing method.
+  <sup>[link](#avoid-bang-methods)</sup>
+
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    def method!
+      # destructive
```
l15n: Maybe _dangerous_ rather than destructive? Destructive operations aren't necessarily dangerous.

ruby/README.md
l15n: > It’s conventional to define the non-bang method in terms of the bang one.

Notably, this is a pretty good rule of thumb. You may wish to incorporate it into the examples

aadityataparia:
@l15n @sikachu I have made suggested changes, please ♻️

ruby/README.md
aadityataparia: @l15n should we add `dependents first` rule in style guide as well?

ruby/README.md
```diff
@@ -60,6 +60,36 @@
...
+
+    ```ruby
+    ## Bad
+    def method!
+      # dangerous operation
+    end
+
+    ## Good
+    def method
+      # dangerous operation
```
davidstosik: This is a bit confusing. Should the comment be clearer, and say something like "dangerous operation, when a safe version of the method is not provided"?

ruby/README.md
```diff
@@ -60,6 +60,36 @@
...
+    end
+
+    ## Good
+    def method
+      # safe operation, preferably in terms of corresponding bang method, i.e.
```
davidstosik: I don't understand what this means. Did you mean "preferably reusing the dangerous method"?

aadityataparia: It means, preferably define safe method in terms of bang method 🤔 https://github.com/cookpad/global-style-guides/pull/104#discussion_r207758641

davidstosik: Sorry, I don't really understand what "in terms of bang method" means. 🙇

ruby/README.md
```diff
@@ -70,22 +70,23 @@
     ```ruby
     ## Bad
     def method!
-      # dangerous operation
+      # dangerous version, without any safe version
```
davidstosik: [Nitpick] _dangerous **operation**, without any safe version_

(Could even add in brackets `(prima donna)`? 🤔 )

ruby/README.md
```diff
@@ -70,22 +70,23 @@
...
     end
 
     ## Good
     def method
-      # dangerous operation
+      # dangerous version, without any safe version
```
davidstosik: [Nitpick] _dangerous **operation**, without any safe version_

PR #105, opened 2018-08-14
Update testing style guides to include rules on the expect to change Rspec syntax

cintamani:
I have recently been fixing a bug on the counter which was incrementing the recipe photo_comment count using `dismissed` new comments, and then re-incrementing when the comment gets `active`. The issue ended up with the photo comment being counted twice in the recipe rather than once. 
The test was there, but since it was using the `expect` `change` Rspec syntax with a `by(1)`, we haven't noticed that the counter which should have been going from 0 to 1, was actually going from 1 to 2. 
Hence, I believe is valuable to add this to our testing style guide

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -415,3 +415,27 @@
     end
     ```
   </details>
+
+- <a name="not-user-expect-change-by"></a>
+  Do not use the `expect` .. `change` syntax with `by`. Use `from` and `to` instead.
+  <sup>[link](#not-user-expect-change-by)</sup>
+
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    it "increment recipe counters" do
+      recipe = create(:recipe)
+       
+      expect{ create(:photo_comment) }.to change{ recipe.reload.photo_comments_count }.by(1)
+    end
+
+    ## Good
+    it "increment recipe counters" do
+      recipe = create(:recipe)
+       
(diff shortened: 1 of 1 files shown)


davidstosik:
Can we please get a link to the test that failed to catch something.

Personally, I think that if you need to expect a change `from(0).to(1)`, then what you actually needed is an expectation of equality *before* the test's exercise, which means you actually needed to test something else? It'd be easier to write and comment on a real example (which I will when I get the link if necessary), but here's my thought process:

```rb
it "increments the counter" do
  # setup

  expect do
    # exercise
  end.to change do
    # fetch counter
  end.by(1)
end
```

This spec above is correct, it is actually testing that whatever we do in the *exercise* part of the test only increments the counter by one.

```rb
it "increments the counter" do
  # setup (including something that accidentally sets the counter to 1)

  expect do
    # exercise
  end.to change do
    # fetch counter
  end.from(0).to(1)
end
```

This spec above not only checks the same thing as the first spec I wrote, it also is a specs that says that something in `# setup` sets the counter to 0. Rather than this spec above, I would rather have two separate specs. First would be the top one in this comment, and the second one would be:

```rb

it "does not increment the counter in {specific scenario}" do
  # setup

  expect do
    # including something that accidentally sets the counter to 1, but should not
  end.not_to change do
    # fetch counter
  end
end
```

The problem with dealing in absolutes is that setup might change over time, and might trigger changes in the database we were not expecting. By making sure we're only testing the exercise, and the outcome in relative changes, we're making our spec more solid.

PR #107, opened 2018-09-11
Enable Lint/Void cop

Knack:
I was playing with RuboCop, forcing it to check offenses that we currently have disabled, and I found this cop particularly interesting because it actually discovered bugs in our specs: https://github.com/cookpad/global-web/pull/9181

I didn't enable the additional option `CheckForMethodsWithNoSideEffects` because it was reporting false positives with Capybara's `select` method.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -280,7 +280,7 @@ Lint/UnderscorePrefixedVariableName:
   Enabled: false
 
 Lint/Void:
-  Enabled: false
+  Enabled: true
 
 Lint/UselessAssignment:
   Exclude:


Knack:
There were some more offenses if I enabled the `CheckForMethodsWithNoSideEffects`. All of them were about using `map` instead of `each`, like:

```ruby
some_scope.map(&:destroy)
```

But then we were getting many false positives about Capybara's `select` (to choose from an HTML `select` tag) because it was confused with `Enumerable#select`

Anyway, I'll fix those offenses too (probably using `destroy_all` in that specific case 🙂 )

PR #108, opened 2018-09-24
Fix example about HTML in translations

Knack:
Just a small fix to the example. The key has to end with '_html' or the `<strong>` tag will be escaped.

https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -280,12 +280,12 @@
     ## Good
     current_time:
       label: "Current time:"
-      label_time: "%{label} %{time}"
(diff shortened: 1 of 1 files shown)


Knack:
> This feels counter-intuitive. 🤔
> There's nothing in label_time that is actual HTML, so I feel weird needing that key to end with _html

@davidstosik In some way, I agree with you

However, this PR was meant to fix the example. I copied the idea in my code and then I realized it didn't work.

As @balvig said, it's the Rails I18n convention 😕

PR #110, opened 2018-12-18
Add another BEM reference

balvig:
I find this documentation to be a really clear and well-written reference for using BEM. 
Thanks to @sikachu for finding it!

One gotcha is that it endorses `block__element_modifier` rather than `block__element--modifier` which we use, so just watch out for that 😓

### DIFF
--- a/scss/bem/README.md
+++ b/scss/bem/README.md
@@ -111,6 +111,7 @@ We try to have as much as possible of:
 
 ## Useful links
 
+- https://en.bem.info/methodology/quick-start/
 - http://fullstackradio.com/episodes/1/
 - https://mattstauffer.co/blog/organizing-css-oocss-smacss-and-bem
 - https://css-tricks.com/bem-101/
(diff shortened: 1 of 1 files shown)


brunobrgs:
Maybe add this `modifier` difference with the link, otherwise people can read and don't know that it is different from what we use.

rikarumi:
I agree on @brunobrgs's point. I think we need to add that gotcha for clearer reference to what we use

balvig:
@brunobrgs @rikarumi any suggestions on how to do that? (feel free to use suggestion feature etc)

scss/bem/README.md
```diff
@@ -111,6 +111,7 @@ We try to have as much as possible of:
 
 ## Useful links
 
+- https://en.bem.info/methodology/quick-start/
```
brunobrgs: Something like: 
```suggestion
- https://en.bem.info/methodology/quick-start Gotcha: ignore `block__element_modifier` and use `block__element--modifier` instead
```

PR #116, opened 2019-01-24
Prefer local variables in partials

sebasoga:
I've been nudging people to follow this convention on code review, as this is a common [_Rails best practice_](http://www.carlosramireziii.com/stop-using-instance-variables-in-partials.html). But as @lewispb [pointed out](https://github.com/cookpad/global-web/pull/10857#discussion_r250244042) having this documented will make it easier to adopt.

Too see how this will look, [click here](https://github.com/cookpad/global-style-guides/blob/b6178ae65bee3d1882ab8aa63450b258eba194e0/rails/README.md#prefer-local-variables-in-partials).

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -353,7 +353,7 @@
   </details>
 
 - <a name="prefer-local-variables-in-partials"></a>
-  Always prefer local variables for a partial over instance variable
+  Always prefer local variables for a partial over instance variables
   <sup>[link](#prefer-local-variables-in-partials)</sup>
 
   <details>
@@ -361,7 +361,11 @@
 
     ```erb
     ## Bad
-    # app/views/users/_follows_count.html.erb
+    
+    <!-- app/views/users/show.html.erb -->
+    <%= render "users/follows_count" %>
+  
+    <!-- app/views/users/_follows_count.html.erb -->
     <div>
       <strong><%= @user.name %></strong><br>
       <span><%= @user.followers.size %> Followers</span>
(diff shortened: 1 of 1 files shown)


rails/README.md
```diff
@@ -351,3 +351,38 @@
...
+    ```erb
+    ## Bad
+    # app/views/users/_follows_count.html.erb
+    <div>
+      <strong><%= @user.name %></strong><br>
+      <span><%= @user.followers.size %> Followers</span>
+      <span><%= @user.followees.size %> Following</span>
+    </div>
+
+    ## Good
```
robertomiranda: Could we list first the rendering part? not sure but at the first sight it looks very similar the bad and good example
```suggestion
    ## Good
   
    # app/views/users/show.html.erb
    <%= render "users/follows_count", user: @user %>
    
    # app/views/me/show.html.erb
    <%= render "users/follows_count", user: current_user %>
    
    # app/views/recipes/show.html.erb
    <%= render "users/follows_count", user: @recipe.user %>    
    
        # app/views/users/_follows_count.html.erb
    <div>
      <strong><%= user.name %></strong><br>
      <span><%= user.followers.size %> Followers</span>
      <span><%= user.followees.size %> Following</span>
    </div>
    ```
```

sebasoga: @robertomiranda I think that's actually the point. Which is to show that the difference seems subtle but the result is quite different as it makes the partial completely reusable, which is what the latter part of the example shows.

davidstosik: What do you think of repeating the bottom partials, and maybe only use one as it's enough to understand the issue?

```erb
<!-- Bad -->

  <!-- app/views/users/show.html.erb -->
  <%= render "users/follows_count" %>

  <!-- app/views/users/_follows_count.html.erb -->
  <div>
    <strong><%= @user.name %></strong><br>
    <span><%= @user.followers.size %> Followers</span>
    <span><%= @user.followees.size %> Following</span>
  </div>


<!-- Good -->

  <!-- app/views/users/show.html.erb -->
  <%= render "users/follows_count", user: @user %>

  <!-- app/views/users/_follows_count.html.erb -->
  <div>
    <strong><%= user.name %></strong><br>
    <span><%= user.followers.size %> Followers</span>
    <span><%= user.followees.size %> Following</span>
  </div>
```

If you really want to, we could add a second caller to illustrate _why_ it is useful to use local variables, but the _why_ is not necessarily this document's purpose? 🤔 

```erb
<!--- Add in good -->
  <!-- app/views/recipes/show.html.erb -->
  <%= render "users/follows_count", user: @recipe.user %>
```

sebasoga: @robertomiranda, @davidstosik thanks for the feedback! I just incorporated it.

sebasoga:
@davidstosik not sure about the indentation but tried to improve it here: https://github.com/cookpad/global-style-guides/pull/119. Feel free to push to the branch directly if you want to change it.

PR #118, opened 2019-01-25
Remove "don't use console" guide

firewalker06:
-


firewalker06:
That also do. Should we also mention that Redshift  is not populated in real time? 

> If you need to export data, I would recommend writing a one-off that writes to a file on S3

Some clarification might be useful like which bucket to use.

PR #120, opened 2019-03-01
Add guide for 'Separate multiline fetch/query methods from the memoize method'

mshka:
A recent [comment on one of my PR](https://github.com/cookpad/global-web/pull/11448#discussion_r261179270) made me realize that this rule was not added in our guide even though there was a discussion about it https://github.com/cookpad/global-web/issues/9372#issuecomment-425327541 I don't think there was a decision about the style but hopefully this would help!

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -241,10 +241,10 @@
 
     ## Good
     def pizza_recipes
-      @_pizza_recipes ||= query_pizza_recipes
+      @_pizza_recipes ||= fetch_pizza_recipes
     end
 
-    def query_pizza_recipes
+    def fetch_pizza_recipes
       Recipe.
         published.
         in_current_language.
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -222,3 +222,33 @@
     end
     ```
   </details>
+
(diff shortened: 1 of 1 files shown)


ruby/README.md
```diff
@@ -222,3 +222,33 @@
...
+
+  <details>
+    <summary>Example</summary>
+
+    ```ruby
+    ## Bad
+    def pizza_recipes
+      @_pizza_recipes ||= Recipe.
+        published.
+        in_provider.
```
balvig:
```suggestion
        in_current_language.
```

(just to make it more universally understandable as this is a public repo 😉 )

ruby/README.md
```diff
@@ -222,3 +222,33 @@
...
+        where("title LIKE ?", "%pizza%")
+    end
+
+    ## Good
+    def pizza_recipes
+      @_pizza_recipes ||= query_pizza_recipes
```
balvig: Super small nit, but just because `query` can appear both as a noun/vern, maybe something like `fetch` sounds _slightly_ better?

```suggestion
      @_pizza_recipes ||= fetch_pizza_recipes
```

PR #125, opened 2019-05-22
Enable "Style/Not" Rubocop rule

davidstosik:
https://github.com/cookpad/global-web/pull/12892#discussion_r286285376


davidstosik:
After testing, it turns out `Style/AndOr` is set to `Enabled: always` by default, so I'll remove it from this PR.

PR #127, opened 2019-05-31
Add explanation to Rails styles

ChrisBr:
Sometimes I'm wondering why we're doing things in a certain way. Although style guides are usually personal preferences, some of our style guides do have some practical reasons which we should document / link. This will make it easier to remember instead of 'just a random rule' and will also help when we hire new (and more junior) team members.

I started with linking to these two good blog articles about [RESTful API design](http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers) and [before_actions](http://craftingruby.com/posts/2015/05/31/dont-use-before-action-to-load-data.html).

WDYT?

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -124,7 +124,7 @@
 
 - <a name="prefer-private-methods"></a>
   Prefer private methods over `before_action` to set instance variables
-  <sup>[link](#prefer-private-methods)</sup>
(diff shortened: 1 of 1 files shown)


ChrisBr:
Cheers 👍

robertomiranda:
awesome 👍

PR #128, opened 2019-06-10
Split .rubocop.yml

davidstosik:
This PR is the first in a series of PRs helping with the [`rubocop`](https://github.com/rubocop-hq/rubocop)/[`rubocop-rails`](https://github.com/rubocop-hq/rubocop-rails) [split](https://github.com/rubocop-hq/rubocop/issues/5976).

Goals:
- when I'm done, we have files in here that can be used by projects with different contexts (for example, a simple Ruby project should not have to require `rubocop-rails` and apply irrelevant Rails rules)
- avoid as much as possible disrupting work on Cookpad projects that currently require `.rubocop.yml`

Next steps:

- introduce `require: rubocop-rails` instruction in `.rubocop.rails.yml`
- (this one is an open question) split [`rubocop-performance`](https://github.com/rubocop-hq/rubocop-performance) the same way?


.rubocop.yml
```diff
@@ -1,328 +1,3 @@
...
-
-Rails/SkipsModelValidations:
-  Enabled: false
-
-Rails/RequestReferer:
-  EnforcedStyle: referrer
-
-Rails/DynamicFindBy:
-  Enabled: false
+inherit_from:
```
davidstosik: I have confirmed that `inherit_from` with relative paths works, at least locally: Rubocop downloaded all required files:
```sh
$ cd global-web/

$ head -2 .rubocop.yml 
inherit_from:
  - https://raw.githubusercontent.com/cookpad/global-style-guides/ds/split-rubocop/.rubocop.yml

$ rm -f .rubocop-*

$ rubocop -c .rubocop.yml

$ ls .rubocop*: ls .rubocop*
.rubocop-https---raw-githubusercontent-com-cookpad-global-style-guides-ds-split-rubocop--rubocop-rails-yml
.rubocop-https---raw-githubusercontent-com-cookpad-global-style-guides-ds-split-rubocop--rubocop-ruby-yml
.rubocop-https---raw-githubusercontent-com-cookpad-global-style-guides-ds-split-rubocop--rubocop-yml
.rubocop.yml
```

lewispb: Does this work with `global-feeds`?

davidstosik: @lewispb As you can read in my goals in the PR description, I want to avoid disrupting any application using the `.rubocop.yml` file.

My theory is that if Rubocop was able to pick the files up, then it will work with all projects pointing to `.rubocop.yml` the way global-web does. 

Are you aware of anything that global-feeds does differently from global-web?

Note that we can as easily revert this PR if we notice any project is broken in the process.

lewispb: Oh I was just wondering if this still worked:

https://github.com/cookpad/global-feeds/blob/master/.rubocop.yml

And if perhaps you'd tested it?

davidstosik: Let me do an additional test using a branch on global-web, for good measure.

davidstosik:
Rubocop did not fail on this global-web pull-request: https://github.com/cookpad/global-web/pull/13196 (note that it [points](https://github.com/cookpad/global-web/pull/13196/files#diff-11a0d7906801d9dea0eccb85667ad811R2) to this PR's .rubocop.yml).

.rubocop.yml
davidstosik: > Oh I was just wondering if this still worked:
> 
> https://github.com/cookpad/global-feeds/blob/master/.rubocop.yml
> 
> And if perhaps you'd tested it?

As per the top comment on this thread:
> I have confirmed that `inherit_from` with relative paths works, at least locally: Rubocop downloaded all required files

Knack:
Looks good! Let me try with `global-newsletters` too.

Should we maybe do the same with the "Performance" cops, since they also were extracted to its own gem `ruby-performance`? 🤔

davidstosik:
@knack
> Performance

That's a question I'm asking in the PR description. :Grimacing:

Knack:
> That's a question I'm asking in the PR description.

🙈 OMG
Well, consider my comment as a **+1** 😉

PR #133, opened 2019-08-21
Avoid writing view specs

davidstosik:
[This recent PR](https://github.com/cookpad/global-web/pull/14483#discussion_r315972634) revived the discussion around view specs. I have confirmed with @balvig on Slack that our stance all these years was to avoid writing view specs. I'll get his help fleshing out this PR with more context and justification when he gets back to work.

Feel free to start a conversation anyway, if you have arguments for or against view specs. 🙏

### DIFF
--- a/rspec/README.md
+++ b/rspec/README.md
@@ -473,3 +473,6 @@
     end
     ```
   </details>
+- <a name="avoid-view-specs"></a>
+  Avoid writing view specs.
+  <sup>[link](#avoid-view-specs)</sup>
--- a/rspec/README.md
+++ b/rspec/README.md
@@ -474,5 +474,5 @@
     ```
   </details>
 - <a name="avoid-view-specs"></a>
-  Avoid writing view specs.
+  Avoid writing view specs, prefer request or feature specs instead.
   <sup>[link](#avoid-view-specs)</sup>


davidstosik:
Here's what I found so far: https://github.com/cookpad/global-web/pull/3668#discussion_r76159783

rspec/README.md
```diff
@@ -473,3 +473,6 @@
     end
     ```
   </details>
+- <a name="avoid-view-specs"></a>
+  Avoid writing view specs.
```
lewispb:
```suggestion
  Avoid writing view specs, prefer request or feature specs instead.
```

balvig:
Will mostly save the discussion for the update to our guides, as we haven't covered this topic 😢  Maybe a conversation from the past can provide a bit of context though:

https://github.com/cookpad/global-web/pull/3668#discussion_r76159783

We can definitely debate whether a view spec can ever be "the right tool for the job".

Ie, is a view spec the best way to verify this particular bug? 

If the answer is "yes", _independent of time constraints or limitations in understanding_, then we can consider tweaking our current convention!

davidstosik:
@sikachu @balvig What should we do with this PR? Do we need to have a debate somewhere before we can decide if view specs are a valuable tool we should add to our toolbox, or less efficient than other kinds of specs that we should stick to?

Or should I merge this and then we can reopen the debate in a future PR discussing the removal of the rule? (Technically, the topic was discussed long ago and the implicit consensus was that we're not using view specs, and this PR is only recording that _de facto_ rule.)

davidstosik:
Thank you for the conclusion @sikachu!

Everyone please keep in mind that no rule is ever final, and we can revisit and discuss what was previously decided. 🙏

PR #123, opened 2019-04-22
Forbid Negated If syntax

davidstosik:
This would detect the _Bad_ examples below:

```rb
# Bad
if !condition
  do_something
end

# Good
unless condition
  do_something
end

# Good
if opposite_condition
  do_something
end

# Bad
do_something if !condition

# Good
do_something unless opposite_condition

# Bad
do_something unless !condition

# Good
do_something if opposite_condition
```

PR #135, opened 2019-08-27
Add Ruby guide regarding unless with operator

ChrisBr:
When using unless, we should avoid using an operator in the conditional. Reasoning behind that the operator also gets negated, leading often to wrong conditionals and harder to understand.

 ```ruby
## Bad
def recipes?
  return false unless foo && bar
end

def recipes?
  return false if !(foo || bar)
end

## Good
def recipes?
  return false unless foo
  return false unless bar
end
```

See https://github.com/cookpad/global-web/pull/14590

PR for a RuboCop https://github.com/rubocop-hq/rubocop/pull/7311

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -252,3 +252,32 @@
     end
     ```
   </details>
+
+- <a name="avoid-unless-with-operator"></a>
+  Avoid `unless` with boolean operator(s) in the conditional
+  <sup>[link](#avoid-unless-with-operator)</sup>
+
+  <details>
+    <summary>Example</summary>
+
+    ```ruby
+    ## Bad
+    unless client_id.present? && client_secret.present?
+      raise "Missing credentials for OauthApplication #{name}."
+    end
+
+    ## Good - flip to if
+    if client_id.blank? || client_secret.blank?
+      raise "Missing credentials for OauthApplication #{name}."
+    end
+
+    ## Good - extract method
+    if missing_credentials?
+      raise "Missing credentials for OauthApplication #{name}."
+    end
+
+    def missing_credentials?
+      client_id.blank? || client_secret.blank?
+    end
+    ```
+  </details>

### CODE REVIEW

File: ruby/README.md [hunk 1]
```diff
@@ -252,3 +252,28 @@
     end
     ```
   </details>
+
+- <a name="avoid-unless-with-operator"></a>
+  Avoid unless with operator in conditional
```
sebasoga: I/m not sure if the correct word I'm looking for is _trailing_ but I think we don't want to go as far as saying that something like this is bad:
```rb
if !(foo || bar)
  false
end
```
```suggestion
  Avoid unless with an operator in trailing conditional
```

ChrisBr: I'm fine with allowing the `if`, I just included the example to show that the operator get also negated in an unless statement which often leads to errors / bad readability. I can include the `if` case in the `good` part tho.

ChrisBr:
> Seems like your PR only covers it for unless though, is the rule for if already defined.

Yes, the cop only catches `unless` as my main point was to not allow `unless` with operators. I'm fine with allowing the negated `if` case though.

sebasoga:
Ah ok, [this example](https://github.com/cookpad/global-style-guides/pull/135/files#diff-9a7d528fdf469b9c5aaa0ae9a9e06d94R270) confused me a bit. Maybe it should be there?

ChrisBr:
> Ah ok, this example confused me a bit. Maybe it should be there?

Yes, I can move it to allowed. I think it is readable, however, personally I would probably prefer to write to unless statements.

image

sebasoga:
> personally I probably write two unless statements instead of a negated if.

I agree and I just noticed I had a typo on my previous comment, I was suggesting to remove (not to move) the example to avoid confusion.

ruby/README.md
davidstosik: @sebasoga Negation in `if` and `unless` conditionals was handled in this PR: https://github.com/cookpad/global-style-guides/pull/123

davidstosik: @sebasoga @ChrisBr FYI, negation in `if` and `unless` conditionals was actually already handled with Rubocop in this PR: https://github.com/cookpad/global-style-guides/pull/123

If I need to write `do_something unless a && b`, but I can't, I will definitely not chose to write `do_something if !(a || b)`. I agree `unless a && b` can be a bit confusing, and I'm fine with not allowing it, but I don't think `if !(a || b)` is acceptable.

If anything, I think complex conditions should be extracted and given an intent-explicit name:
```rb
## Bad
def action
  do_something unless a && b

  # ...
end

## Better
def action
  do_something 

  # ...
end

def do_something?
  !(a && b)
end
```

davidstosik: Nit:
```suggestion
  Avoid `unless` with boolean operator(s) in the conditional
```

ruby/README.md
```diff
@@ -252,3 +252,28 @@
...
+  <sup>[link](#avoid-unless-with-operator)</sup>
+
+  <details>
+    <summary>Example</summary>
+
+    ```ruby
+    ## Bad
+    def recipes?
+      return false unless foo && bar
```
davidstosik: [Nit] I think it would help drive the point if we could come up with real-life examples and names.

For example (from our code-base, but simplified):
```rb
## Bad
def captcha_verified?
  return true unless signing_up_from_web? && signing_up_with_email?

  captcha.valid?
end

## Good
def captcha_verified?
  return true unless signing_up_from_web?
  return true unless signing_up_with_email?

  captcha.valid?
end

## Also good
def captcha_verified?
  return true unless should_verify_captcha?

  captcha.valid?
end

def should_verify_captcha?
  signing_up_from_web? && signing_up_with_email?
end
```

ChrisBr:
Thanks @balvig and @davidstosik 👍 Agreed!

Any good arguments to convince RuboCop HQ tho? (see PR in inital comment)

ChrisBr:
♻️

davidstosik:
For the record:

> Oh by the way, not all the rules we want to apply with Rubocop have to be in Rubocop's codebase (https://github.com/cookpad/global-web/pull/13244)

PR #136, opened 2019-08-30
Avoid jQuery

davidstosik:
Hoping I can officialize this: https://github.com/cookpad/global-web/issues/9333#issuecomment-434846695

Not sure it needs an example though..? 🤔

### DIFF
--- a/javascript/README.md
+++ b/javascript/README.md
@@ -1,5 +1,22 @@
 # Javascript
 
+- <a name="avoid-jquery"></a>
+  Avoid using jQuery and jQuery plugins. Prefer solutions based on plain Javascript or Stimulus.
+  <sup>[link](#avoid-jquery)</sup>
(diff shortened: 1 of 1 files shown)


brunobrgs:
WDYT about moving the points listed in the issue to the description here?

> - Don't add any new JS code that uses jQuery
> - If you're changing a piece of code that uses jQuery, change it to not use jQuery
> - If you find a piece of code where changing it not to use jQuery is not straightforward (i.e. plug-ins), open an issue on global-web with the Kaizen tag so that we can do that later

So we can see more details straghtforward, instead of having to open the link.
Are these points all appliable as a rule?

davidstosik:
@sebasoga Note that the _explanations_ link points to [your original comment](https://github.com/cookpad/global-web/issues/9333#issuecomment-434846695) including the details. 😉

PR #143, opened 2019-10-18
Make BEM "bible" link more prominent.

jesseclark:
This PR adds some of the information/advice that came up in discussion over on #15315.

Additionally, I paraphrased the answer to the 'Why BEM?' question from [this article](https://blog.decaf.de/2015/06/24/why-bem-in-a-nutshell/). I have found this answer to be the most concise and relevant of all those that I have come across so far.


jesseclark:
cc: @cookpad/web-chapter

scss/bem/README.md
```diff
@@ -5,7 +5,10 @@ Currently not all CSS have been converted to this style so it's a work in progre
 
 ## BEM
 
-BEM is a guideline for naming classes that prevent mistakenly overriding other classes:
+BEM is a set of guidelines for naming classes which helps us produce maintainable CSS by:
+- Avoiding the infinite inheritance of CSS and providing the concept of clojures by using unique CSS classes per element <sup>[1](useful-links-and-citations)</sup>
```
sebasoga: I think there's a typo:
```suggestion
- Avoiding the infinite inheritance of CSS and providing the concept of closures by using unique CSS classes per element <sup>[1](useful-links-and-citations)</sup>
```

scss/bem/README.md
```diff
@@ -15,6 +18,8 @@ BEM is a guideline for naming classes that prevent mistakenly overriding other c
 - **Element:** Parts of a UI component
 - **Modifier:** Different states/exceptions
 
+We follow the recommendations of https://en.bem.info/methodology/quick-start as our BEM style guide with one exception: 
+- instead of `block__element_modifier` we use `block__element--modifier`
```
sebasoga: Nitpick:
```suggestion
- Instead of `block__element_modifier` we use `block__element--modifier`
```

scss/bem/README.md
```diff
@@ -104,14 +109,17 @@ We try to have as much as possible of:
 - "Form with chromeless inputs"
 - "Hint text"
 
+However, when it isn't clear if a section should be a block or an element or if a block is a candidate for reuse, it is ok to start specific and if reuse becomes possible later on, we can refactor to make things more generic.
```
ollieh-m: 👍

scss/bem/README.md
```diff
@@ -15,6 +18,8 @@ BEM is a guideline for naming classes that prevent mistakenly overriding other c
 - **Element:** Parts of a UI component
 - **Modifier:** Different states/exceptions
 
+We follow the recommendations of https://en.bem.info/methodology/quick-start as our BEM style guide with one exception: 
+- instead of `block__element_modifier` we use `block__element--modifier`
 
 ### Example
```
ollieh-m: While we're updating this file... Should we change the margin in this example:

```
.tab-list {
  margin: 10px 0;

  &__item {
    display: inline-block;

    &--active {
      background: #EEE;
    }
  }
}
```

To be:

```
.tab-list {
  padding: 10px 0;

  &__item {
    display: inline-block;

    &--active {
      background: #EEE;
    }
  }
}
```

balvig: Good point 👌

scss/bem/README.md
jesseclark: Haha. Yep. I spelled it like the language instead the feature.

scss/bem/README.md
jesseclark: 👍  Done.

sebasoga: I'm a bit lost here, is there something I can read to understand why we prefer using `padding` instead of `margin`?

ollieh-m: The following comes from https://en.bem.info/methodology/quick-start/:

> The block shouldn't influence its environment, meaning you shouldn't set the external geometry (margin) or positioning for the block.

The idea I think is that each css block should be reusable in any context. Suppose I want the tab-list 50px from the bottom of a footer, and then elsewhere I want it 10px from the top of a header. If the block itself sets its own margin that becomes awkward. But if it only styles the _content_ of a rectangular space, that rectangle can be put anywhere. Some parent block (`footer` or `header`, say) can then determine _where_ the tab-list is positioned, like this:
```
.footer { padding: 50px }
.header { padding: 10px }

<div class="header">
  <div class="tab-list"></div>
</div>

<div class="footer">
  <div class="tab-list"></div>
</div>
```

sebasoga: Thanks @ollieh-m, that makes sense!

davidstosik:
Thank you for this update! 👍 
I opened a PR to fix the anchor links: #144.

PR #145, opened 2019-10-23
Remove numbering typo

firewalker06:
This additional numbering is not needed in this unordered list.

### DIFF
--- a/scss/bem/README.md
+++ b/scss/bem/README.md
@@ -118,7 +118,7 @@ However, when it isn't clear if a section should be a block or an element or if 
 
 ## Useful Links and Citations
 
-- 1: https://blog.decaf.de/2015/06/24/why-bem-in-a-nutshell/
+- https://blog.decaf.de/2015/06/24/why-bem-in-a-nutshell/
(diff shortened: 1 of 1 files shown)


firewalker06:
> My intention was for the `1:` to be the citation/footnote number from the link on line 9: `<sup>[1](useful-links-and-citations)</sup>`. Removing this would make the citation confusing.

Ah, now I understand how it works. Markdown isn't a good format to have proper citation like in Wikipedia 😅.

> Maybe we should make 'Useful Links' and 'Citations' two separate sections and have 'Citations' contain an ordered list? We could even rename 'Citations' to 'References' which is the terminology the Wikipedia uses for their footnotes/citations section.

Either this, or make the list into numbered list of links:

```
1. Why BEM in a Nutshell
2. Using BEM to solve common issues in web development
3. Matt Stauffer - CSS Semantics
4. BEM 101
```

Btw, the link for Fullstackradio is broken and should be updated to `http://www.fullstackradio.com/1`.

firewalker06:
Closing this and prefer #146

PR #146, opened 2019-10-24
Change links and citations to ordered list.

jesseclark:
Also improved link text and fixed broken link.

Based off of a proposal from @firewalker06 in #145

### DIFF
--- a/scss/bem/README.md
+++ b/scss/bem/README.md
@@ -18,7 +18,7 @@ BEM is a set of guidelines for naming classes which helps us produce maintainabl
 - **Element:** Parts of a UI component
 - **Modifier:** Different states/exceptions
 
-We follow the recommendations of https://en.bem.info/methodology/quick-start as our BEM style guide with one exception: 
+We follow the recommendations of the [Official BEM Methodology](https://en.bem.info/methodology/quick-start) as our BEM style guide with one exception: 
 - Instead of `block__element_modifier` we use `block__element--modifier`
 
 ### Example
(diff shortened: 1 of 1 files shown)


scss/bem/README.md
```diff
@@ -118,8 +118,8 @@ However, when it isn't clear if a section should be a block or an element or if
 
 ## Useful Links and Citations
 
-- 1: https://blog.decaf.de/2015/06/24/why-bem-in-a-nutshell/
-- https://en.bem.info/methodology/solved-problems/
-- http://fullstackradio.com/episodes/1/
-- https://mattstauffer.co/blog/organizing-css-oocss-smacss-and-bem
-- https://css-tricks.com/bem-101/
+1. [Why BEM In A Nutshell](https://blog.decaf.de/2015/06/24/why-bem-in-a-nutshell/)
+1. [BEM Methodology - Solved Problems](https://en.bem.info/methodology/solved-problems/)
```
davidstosik: I've never seen that before, but using `1.` for all numbered entries is interesting...

I understand that it avoids having to maintain proper numbering and ordering, while formatted text will show a properly numbered list, and I like that, but I also consider Markdown as a plain-text format that is readable as-is (even in Vim if I'd like to).
Not numbering this list makes the plain-text format less readable, and in my opinion, goes against Markdown's purpose.

> The overriding design goal for Markdown’s formatting syntax is to make it as readable as possible. The idea is that a Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions. While Markdown’s syntax has been influenced by several existing text-to-HTML filters, the single biggest source of inspiration for Markdown’s syntax is the format of plain text email. _([source](https://daringfireball.net/projects/markdown/))_

I appreciate the readability improvements this PR brings, and I would have approved it as-is, but I'd be curious to hear your arguments for keeping the "all 1 ordered list".  🙇

scss/bem/README.md
```diff
@@ -118,8 +118,8 @@ However, when it isn't clear if a section should be a block or an element or if
...
+1. [Full Stack Radio - Matt Stauffer - CSS Semantics](http://www.fullstackradio.com/1)
+1. [Organizing CSS OOCSS SMACSS And BEM](https://mattstauffer.co/blog/organizing-css-oocss-smacss-and-bem)
+1. [BEM 101](https://css-tricks.com/bem-101/)
```
davidstosik: @jesseclark 
I've never seen that before, but using `1.` for all numbered entries is interesting...

I understand that it avoids having to maintain proper numbering and ordering, while formatted text will show a properly numbered list, and I like that, but I also consider Markdown as a plain-text format that is readable as-is (even in Vim if I'd like to).
Not numbering this list makes the plain-text format less readable, and in my opinion, goes against Markdown's purpose.

> The overriding design goal for Markdown’s formatting syntax is to make it as readable as possible. The idea is that a Markdown-formatted document should be publishable as-is, as plain text, without looking like it’s been marked up with tags or formatting instructions. While Markdown’s syntax has been influenced by several existing text-to-HTML filters, the single biggest source of inspiration for Markdown’s syntax is the format of plain text email. _([source](https://daringfireball.net/projects/markdown/))_

I appreciate the readability improvements this PR brings, and I would have approved it as-is, but I'd be curious to hear your arguments for keeping the "all 1 ordered list".  🙇

jesseclark:
I was unaware that plain text formatting was Markdown's overriding purpose. 

Since this document is hosted on Github and gets automatically converted to HTML and that is how users will interact with it the majority of the time, I was following [Github's Markdown Syntax](https://guides.github.com/features/mastering-markdown/) suggestions.

Happy to change it though.

PR #149, opened 2019-11-06
add explanation about JavaScript and progressive enhancement

balvig:
As I have several times in the past failed to explain this concept within the scope of pull requests, I set some time aside to write-up a blog post.

https://sourcediving.com/progressive-enhancement-as-a-valuable-philosophy-in-the-age-of-javascript-aac2e26364d2

This approach has been misinterpreted in a number of ways, including "absolutely everything must work with JavaScript off", so hopefully the post clarifies the sweet spot.

Feel free to ask any questions here as well, would love to discuss it in further detail!

PR #150, opened 2019-11-12
Add reference to 'Dealing with flaky tests' blog article

davidstosik:
In the same vein as @balvig's #149, after a few PR reviews that ended on the long side, and a few flaky specs I worked on myself, I wrote [dealing with flaky tests](https://sourcediving.com/dealing-with-flaky-tests-d7c17227ab26) to act as a reference I could refer to when dealing with a flaky spec myself, or reviewing a pull request for which it could be helpful.

I'll admit this is not really a style guide, but this is a methodology (good practice?) that I believe could help in many instances of trying to fix a flaky spec.

I'll be more than happy to discuss details of the article, or even amend it if anyone has valuable suggestions! 🙏

### DIFF
--- a/rspec/README.md
+++ b/rspec/README.md
@@ -487,4 +487,4 @@ These are some of the conventions we follow:
   </details>
 - <a name="#methodical-flaky-specs"></a>
   Be methodical towards flaky specs.
-  <sup>[link](#methodical-flaky-specs) [explanation](https://sourcediving.com/dealing-with-flaky-tests-d7c17227ab26)</sup>
+  <sup>[link](#methodical-flaky-specs) [explanation](https://medium.com/@davidstosik/a-methodological-approach-to-fixing-flaky-tests-92a39162b769?postPublishedType=repub)</sup>
--- a/rspec/README.md
+++ b/rspec/README.md
@@ -485,3 +485,6 @@ These are some of the conventions we follow:
     end
     ```
   </details>
+- <a name="#methodical-flaky-specs"></a>
+  Be methodical towards flaky specs.
+  <sup>[link](#methodical-flaky-specs) [explanation](https://sourcediving.com/dealing-with-flaky-tests-d7c17227ab26)</sup>


ChrisBr:
Thanks for this very detailed explanation @davidstosik 👏 Here are some small suggestions.

1) Can we use GitHub in advance for a review of blog articles. I think it is kinda hard to make suggestions etc. when it is already published on sourcediving.com. So I would propose we store all blog posts additionally in some repository and follow the PR approach. Something like `web-chapter/blog/`? cc @balvig 

2) The blog article is very very detailed which is great. However, personally I think it would've benefited from splitting into several blog articles. Maybe `Roots of flaky specs` and `How to fix flaky specs`. I don't think it makes sense to split it now after it already got published but maybe something to keep in mind for the future?

3) I'm sure you already know about this feature but CircleCI also has an insight tabs which will show specs who recently failed. Maybe worth to mention too as you referenced several times to CircleCI?
image

4) The idea of creating `RSpec::OrderedCommandFormatter` gem is 👍 

Anyway, thanks for writing it down!

tejanium:
@ChrisBr PR for the blog was here https://github.com/cookpad/developer-blog/pull/56

rspec/README.md
```diff
@@ -485,3 +485,6 @@ These are some of the conventions we follow:
     end
     ```
   </details>
+- <a name="#methodical-flaky-specs"></a>
+  Be methodical towards flaky specs.
```
balvig: Just an idea, but:

```suggestion
  Don't ignore flaky specs.
```

or

```suggestion
  Address flaky specs when they appear.
```

balvig:
> However, personally I think it would've benefited from splitting into several blog articles. 
> Maybe Roots of flaky specs and How to fix flaky specs. 
> I don't think it makes sense to split it now after it already got published

Depending on what the result looks like, that actually sounds like a pretty good idea to me, even post-publication? What do you think @davidstosik?

> Can we use GitHub in advance for a review of blog articles.

@ChrisBr to be honest, the flow for reviewing blog posts is a little broken at the moment: https://github.com/cookpad/developer-blog/issues/54

Namely, there is no easy flow from `.md` file to Medium, and I've always found I end up editing the blog quite a bit once I see it on there anyway, further complicating the review flow.

That's also why I "broke the rules" for my [last blog](https://medium.com/@balvig/progressive-enhancement-as-a-useful-philosophy-in-the-age-of-javascript-aac2e26364d2) by going straight to Medium. 👮‍♂️🙇

This time around I asked for manual feedback from a couple of people using the Medium "private comment" feature + Slack before submitting and ultimately getting review from Miles, the "owner" of the blog.

Your excellent suggestion being missed for this post is a good case in point as to why it could be worth at least opening up a wider net for feedback though, whether using markdown files or Medium. 🤔 

Probably requesting review from @cookpad/web-chapter would have worked this time, then ,aybe for "direct to Medium"-posts, we can simply commit a link to them or something? 🤔

davidstosik:
@ChrisBr Thank you for taking the time to read my post and to make suggestions!

1. Indeed, I guess our current process around blogs and referring blogs into our documentation might need some clarification and improvement. We've indeed had https://github.com/cookpad/developer-blog for a long time. My mistake is probably that I should have requested a review from the whole chapter there. Then introducing a link to that blog post in our documentation would have been nothing controversial.

2. Yeah, I guess I could split this blog post in two... I'm not sure I want to spend the extra time required on formatting and reorganizing: each blog post is gonna need at least its intro and conclusion, right? Then referring to the earlier part of the post from the second part is going to be harder, and assume that the reader already the other blog post... What would you recommend?

3. Not sure that screen is super relevant to flaky specs. 🤔 To be honest, I've never felt the need to look at that screen to track down flaky specs I was not aware of. Flaky specs are enough annoyance that just showing up should trigger an action from the developer. Moreover, CircleCI's insights screen feels kinda broken (half of the screen is currently useless as we're using workflows), impractical (the screen is relative to a given branch) and quite cryptical (what exactly is listed in "failed tests"? How recent are those failures, and why can't I even go back to the jobs where those failures happened?). Considering it's never been useful to me, I didn't see any good reason to add it. If you have a practical example of how that can be useful, I'll welcome it and probably find the time to add it to my blog post. :)

4. Thank you!

davidstosik:
Felt like I had to give it a try:

- Original post: [Dealing with flaky tests](https://medium.com/p/d7c17227ab26)
- First half: [Flaky Tests are NOT Random Failures!](https://medium.com/p/363fbddc1774)
- Second half: [A methodological approach to fixing flaky tests](https://medium.com/p/92a39162b769)

Opinions? (Also, does anyone have a suggestion of image for the second half? Something better than an insect under a magnifying glass...)

ChrisBr:
> Felt like I had to give it a try:

It's great 👍 I like it!

Not sure if we already announced it somewhere (e.g. Twitter) but now you could even publish the two posts in batches. One today and then announce the second post in a week. Two times noise - two times win! You could write something like `Stay tuned for the second part of the blog post which will get released next week.`

## Image

I also don't really understand what this image means nor what it has to do with flaky speaks (except there is flake mentioned on the chocolate bar?)
image

What about something like some person sitting desperate in front of a computer? And then mention flaky specs are nothing to get desperate about because there are methodical approaches to fix them. (just quick google search without proper license).

image

Or some snake / turtle because flaky specs slow down the development team?

image

Just some random ideas ...

balvig:
Ditto what @ChrisBr said, splitting it in two works really well, nice one giving it a shot! 👌 

I also like the snail photo for whatever reason. 😆 

giphy

If I were to make just one subjective suggestion, it might be to make the first title a little less shouty, maybe simply `Flaky tests are not random`?

davidstosik:
> If I were to make just one subjective suggestion, it might be to make the first title a little less shouty, maybe simply `Flaky tests are not random`?

Done.

> I also don't really understand what this image means nor what it has to do with flaky speaks (except there is flake mentioned on the chocolate bar?)

Yeah, the Flake is flaky chocolate, nothing else. I'm fine with using the snails for the second part, but somehow, I'd feel unhappy with starting a post that's supposed to give advice and to unstuck people, with an image that has a negative connotation. Sure, flaky specs slow the dev team down, but this is not the point of these blog posts (I even [removed](https://github.com/cookpad/developer-blog/pull/56/commits/6c972fb91db668a66ddf6e24cdc9a8941f79d323) a big part of my post that was going in details in the detriments that flaky specs bring.) On the contrary, I'd prefer to see an image that brings positivity, shows that there's hope behind the seemingly _unsolvable random failure_.

> Two times noise.

Yeah, not sure about this one either... 😕 
This is not really a marketing piece, this is knowledge that I've intended to share with the team first, then thought it could benefit other people too so I turned it into a blog post after delaying it for too long, and I don't see any very compelling reason to delay it any longer...?

balvig:
@davidstosik whatever image / publishing strategy works for you works for me 😉

ChrisBr:
> On the contrary, I'd prefer to see an image that brings positivity, shows that there's hope behind the seemingly unsolvable random failure.

That makes sense! These were just some random suggestions, feel free to ignore them.

> This is not really a marketing piece,

Yeah, I get that the original intention was just for internal knowledge sharing. However, this is really really great stuff so I think we should take advantage of this! In my opinion it is too good to be only shared for 'internal use'.

> whatever image / publishing strategy works for you works for me 

Same for me, just some random suggestions from my side but I think it is good to go as it is!

Thanks again for writing this down!

davidstosik:
Sorry for the delay. I'll just go with the snails. 🙇

PR #151, opened 2019-11-25
Add @web-chapter to CODEOWNERS

balvig:
As with https://github.com/cookpad/web-chapter/pull/173

Adds @cookpad/web-chapter to CODEOWNERS to leverage GitHub automation.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,1 @@
+* @cookpad/web-chapter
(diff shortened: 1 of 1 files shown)


sebasoga:
Given that in theory the purpose of this repository is to have any Cookpad engineering team to share their styles guides here (ML uses the [phyton/](https://github.com/cookpad/global-style-guides/tree/master/python) folder). Should this be a bit more granular?

balvig:
> Should this be a bit more granular?

🤦‍♂ ...that thought had crossed my mind earlier and then somehow completely forgot about it. 🙇 

Yes! Good point, we don't need to look at `/python` f.ex 👌

balvig:
♻️ please 🙏

PR #153, opened 2020-01-16
Add explanation to 'Dont add HTML to locale file'

ChrisBr:
I still frequently see that we add `_html` keys to our locale files.  @davidstosik wrote a good How To in one of our recent translation PRs. 

I'm planning to summarise this in a blog article but as long as we don't have this, we should add a link to this comment so it does not get lost and add some more context.

https://github.com/cookpad/global-web/pull/15763#issuecomment-554836409

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -261,8 +261,8 @@
 
 - <a name="dont-html-in-locale"></a>
   Don't include HTML in locale file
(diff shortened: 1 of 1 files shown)


davidstosik:
> > planning to summarise this in a blog article
>
> Terrific! 👌

Ooh, I had missed that! _Terrific_ indeed! 👏

PR #157, opened 2020-03-06
Add preference on `click_on` into the style guide

sikachu:
As @balvig pointed me out multiple times during my task of migrating old specs, we now prefer `click_on` over a specific `click_link` or `click_button` in feature specs.

This PR is to make it an official preference into the style guide, <del>and hopefully we can write a cop to automatically check for this</del>.

/cc @cookpad/web-chapter


sikachu:
Since this got a universal thumbs-up, and looks like there's no objection, I'm going to merge this in.

PR #159, opened 2020-03-12
replace SCSS guides with links to TailwindCSS and BEM

balvig:
Sibling of https://github.com/cookpad/global-web/pull/17568.

Quoting:

> Using TailwindCSS and HTML components, needing to think about BEM becomes less common, but we can however still use it for extracting CSS components.

> With this combination, we have the docs from both [TailwindCSS](https://tailwindcss.com/) and [BEM](https://en.bem.info/methodology/quick-start/) to rely on, and so although I can imagine new conventions taking form over time, I don't think we need our current style guides anymore.

> I'm thinking therefore to remove them for now, and instead defer to the official docs?


balvig:
> What do you think about copying over the useful links as well? I think those might still be useful for people who getting started in using BEM.

@sikachu Ah, to be honest I think the official guide has now gotten so good, that to the extent we're using it, it should be sufficient :)

balvig:
> Nice! Might become a bit harder to link to specific rules, but not a big deal?

Yeh, the rules we had actually don't even really apply anymore, but I wouldn't be surprised if we will need to add some again once we get going :)

PR #160, opened 2020-03-16
Dont assert magic values

ChrisBr:
This is something we already practice but with new people joining the team it's good to have it written down. 

```ruby
## Bad
it "does something" do
  user = create(:recipe)

  expect(recipe.title).to eq("Recipe Title")
 end

## Good
it "does something" do
  user = create(:recipe, title: "Recipe Title")

  expect(recipe.title).to eq("Recipe Title")
end
```

With `Avoid adding dummy data that is not asserted in the spec` we already have something similar. Maybe we should merge these two?

### DIFF
--- a/rspec/README.md
+++ b/rspec/README.md
@@ -492,3 +492,26 @@ These are some of the conventions we follow:
 - <a name="#methodical-flaky-specs"></a>
   Be methodical towards flaky specs.
   <sup>[link](#methodical-flaky-specs) [explanation](https://medium.com/@davidstosik/a-methodological-approach-to-fixing-flaky-tests-92a39162b769?postPublishedType=repub)</sup>
+- <a name="#magic-guest"></a>
+  Explicitly setup values used in assertions.
+  <sup>[link](#magic-guest) [explanation](https://thoughtbot.com/blog/mystery-guest)</sup>
+
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    it "does something" do
+      user = create(:recipe)
+
+      expect(recipe.title).to eq("Recipe Title")
+    end
+
+    ## Good
+    it "does something" do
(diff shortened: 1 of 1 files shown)


rspec/README.md
```diff
@@ -492,3 +492,26 @@ These are some of the conventions we follow:
 - <a name="#methodical-flaky-specs"></a>
   Be methodical towards flaky specs.
   <sup>[link](#methodical-flaky-specs) [explanation](https://medium.com/@davidstosik/a-methodological-approach-to-fixing-flaky-tests-92a39162b769?postPublishedType=repub)</sup>
+- <a name="magic-value"></a>
+  Don't assert / expect magic values.
```
balvig: I think a lost a comment somewhere, but made the suggestion that maybe we could define "magic values" a little better, either:

```suggestion
  Don't assert values whose origin is unclear.
```

_or_, I don't know how well-known the concept of ["mystery guests"](https://www.quora.com/What-is-Mystery-Guest-in-TDD) is, but could also be:

```suggestion
  Don't make assertions on "mystery guests".
```

sikachu: Maybe we could say something along the line of "Don't make assertion on the value that you didn't set up in the setup phrase?" ... not sure how to make it concise though.

jesseclark: Maybe it works better to suggest what we should do instead of what we should not do? Or both?
```suggestion
  Don't assert / expect magic values.
  Explicitly setup values used in assertions (e.g. don't assert on unsepecified factory defaults)
```

balvig: @jesseclark Fantastic 👍 I think it works really well even without the parens (the code examples I think serve as the `e.g.` part in this case)

ChrisBr:
Brilliant, thanks all! Updated and added a link to the Thoughtbot blog article.

https://thoughtbot.com/blog/mystery-guest

PR #161, opened 2020-03-27
Add enum best practice

ChrisBr:
Triggered by the discussion in https://github.com/cookpad/global-web/pull/17851#discussion_r399356551, I was wondering what we prefer to use to store an enum field.
In our codebase, we currently use both integers and strings. I'm not suggesting that we should refactor however we decide but when we introduce a new enum we should follow the style guide.

## String
Using a string as enum in the database makes the data more descriptive without the codebase. Plain SQL queries will be more descriptive as they don't contain numbers but strings instead.

## Integers
We can potentially change easily the name of an enum state without the need of a migration as the integer value will stay the same.

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -380,3 +380,25 @@
   Leverage top-down development and feature toggles to keep pull requests small.
   <sup>[link](#leverage-top-down)
 [explanation](https://sourcediving.com/a-practical-guide-to-small-and-easy-to-review-pull-requests-a7f04a01d5d5)</sup>
+- <a name="enum"></a>
+  Use string to store enum state instead of integer.
+  <sup>[link](#enum)</sup>
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    # Bad
+    class AddColumnMigration < ActiveRecord::Migration[6.0]
+      def change
+        add_column :status, :table, :integer
+      end
+    end
+
+    # Good
+    class AddColumnMigration < ActiveRecord::Migration[6.0]
+      def change
+        add_column :status, :table, :string
+      end
+    end
+    ```
+  </details>


davidstosik:
+1 for using integers.
I actually like the fact that using integers obfuscates the database, as it prevents external entities like the data team to rely too much on implementation details that we might change without warning.

jesseclark:
> I actually like the fact that using integers obfuscates the database, as it prevents external entities like the data team to rely too much on implementation details that we might change without warning.

I come to the opposite conclusion for similar reasons. 😁 

I agree with Christian's points [in his original comment on the other PR](https://github.com/cookpad/global-web/pull/17851#discussion_r399356551)

Storing a string value instead of an integer makes it easier to construct, debug, and interpret the results of queries. In some cases the data team's reporting could be relying on the interpretation of the values of the enum staying consistent and we probably shouldn't be changing them in code without considering the downstream impacts and communicating with other teams. Storing string values prevents other teams from having to create lookup tables that map integers to the string values in our code which could then get out of sync without warning.

balvig:
> In some cases the data team's reporting could be relying on the interpretation of the values of the enum staying consistent 

@jesseclark ah, just to clarify on this note: Data teams relying directly on master tables is  "deprecated" (f.ex [here](https://ckpd.slack.com/archives/C7G0A0ML5/p1560910101035000?thread_ts=1560761801.027300&cid=C7G0A0ML5), and [here](https://github.com/cookpad/global-inbox/issues/836)), and the expectation is that they mainly rely on activity logs or events which are immutable, and will ultimately get to a place where no tables are referenced directly.

In the meantime we aim to be pragmatic and support where we can, but I think we should avoid this determining how we write code meant to drive the _behavior_ of the site, as we could end up being unable to make changes due to invisible dependencies outside of the app.

jesseclark:
> > In some cases the data team's reporting could be relying on the interpretation of the values of the enum staying consistent
> 
> @jesseclark ah, just to clarify on this note: Data teams relying directly on master tables is "deprecated" (f.ex [here](https://ckpd.slack.com/archives/C7G0A0ML5/p1560910101035000?thread_ts=1560761801.027300&cid=C7G0A0ML5), and [here](https://github.com/cookpad/global-inbox/issues/836)), and the expectation is that they mainly rely on activity logs or events which are immutable, and will ultimately get to a place where no tables are referenced directly.
> 
> In the meantime we aim to be pragmatic and support where we can, but I think we should avoid this determining how we write code meant to drive the _behavior_ of the site, as we could end up being unable to make changes due to invisible dependencies outside of the app.

Fair enough on not coupling external systems to our operational database. However, in those activity logs or events that we would sending to other systems would we be sending things like `status: 3` or `status: "pending"`?

I would rather optimize for human readibility throughout the stack including the database, personally. I think that debugging SQL code or database values might be a more common use case than redefining an enum value which may trigger the need for a migration if that enum were stored in the db as a string. If we find situations where doing string comparisions in the database were hurting performance that could be optimized on case-by-case basis.

I don't feel super strongly about it since at least Rails console will convert enum values into their string equivalents when inspecting objects, but strings in the db is my preference.

balvig:
> in those activity logs or events that we would sending to other systems would we be sending things like status: 3 or status: "pending"?

It would be up to the data team to make that request I guess, and we would write specs around it to enforce the contract.

> I don't feel super strongly about it 

Same for me to be honest. 

That is where I think considering that the vast majority of existing enums are integers, and debugging SQL historically doesn't seem to have been a major issue for us, if we want to create a guide around it I think it's worth valuing consistency in this case.

jesseclark:
> That is where I think considering that the vast majority of existing enums are integers, and debugging SQL historically doesn't seem to have been a major issue for us, if we want to create a guide around it, I think it's worth valuing consistency in this case.

I think I would prefer not to have a recommendation and allow both then. 

TBH, I don't really like the way AR enums try to combine both enumeration functionality and a barebones state machine. I think it kinda ends up doing both somewhat poorly (more so the state machine bit, to be fair). I've used the enumerize gem on a project in the past and found it to be a bit better than AR's enum.

I came across a situation yesterday where it seemed like there is a legitimate (i.e. beyond preferring human readability) reason for wanting to keep the values of an enum as strings: https://github.com/cookpad/global-web/pull/17944. Granted, I could stick with constants and achieve the desired result of not referencing a magic string, but an enum seems like a somewhat cleaner solution. Curious what others thoughts are.

balvig:
> I came across a situation yesterday where it seemed like there is a legitimate (i.e. beyond preferring human readability) reason for wanting to keep the values of an enum as strings

@jesseclark I think that was a great refactoring 👍 and even if we did have "prefer integer enums" as a guide, I don't think we would let it come in the way of an improvement like that.

I guess also in this case we were essentially dealing with an _existing_ string enum, so a guide that suggested using integers when introducing _new_ enums wouldn't necessarily work against it.

Incidentally, I came across [a recent example showing the refactoring advantage for integers](https://github.com/cookpad/global-web/pull/18217) as well. 

Had it been a string, I suppose it would still have been possible to do either:

- `:only_new_recipes => :rejecting_preexisting_recipes` (maybe confusing?)
- somehow run a migration (the async nature would maybe make it tricky?)

---

As mentioned, if I was asked to write a guide now, I would lean towards favoring integers.

I haven't personally experienced many situations where SQL readability made string enums preferable (at least not with Rails! 😅), and the official word in https://github.com/rails/rails/pull/26422#issuecomment-245349731 is a bit unsettling (anyone able to find a more recent source?)

However, as also mentioned it's not something I feel strongly about. We haven't seen any major problems with either approach and it's relatively easy to change both ways.

I think guides can be used in multiple ways, anywhere from removing the mental overhead of decisions that don't _really_ matter, to taking a firm stance on things that are very important to us.

This one however seems to fall somewhere in between, with no overwhelmingly clear pros/cons (as far as I've seen in _current_ experience), so barring any new information to this thread to sway it in either direction, I would also be fine with "let's wait a bit more and see". 🤔

ChrisBr:
Thank @balvig! Closing the PR for now.

PR #165, opened 2020-05-13
Expanded description of "don't abuse the zero key"

jesseclark:
I think the information about pluralization keys being configured to be automatically added and dropped by OneSky is useful for understanding the "don't abuse the zero key" rule.

Inspired by this discussion in Slack: https://ckpd.slack.com/archives/C06B9GD9D/p1589364941475900

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -256,8 +256,19 @@
   <sup>[link](#scope-generic-phrases-under-common)</sup>
 
 - <a name="dont-abuse-zero-key"></a>
-  Don't use i18n `zero:` key to display a "no results" message (it is intended only to allow proper grammar)
+  Don't use i18n `zero:` key to display a "no results" message. 
   <sup>[link](#dont-abuse-zero-key)</sup>
+  <details>
+    <summary><em>Explanation</em></summary>
+  
+    <pre>  
+      Pluralization rules vary from language to language and keys are automatically added and 
+      removed from the translation files. So, the `zero:` key cannot be relied on to be present 
+      for every language. 
+      
+      Use a specific key for the "no results" message instead.
+    </pre>
+  </details>
 
 - <a name="dont-html-in-locale"></a>
   Don't include HTML in locale file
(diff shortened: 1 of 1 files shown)


rails/README.md
```diff
@@ -256,8 +256,19 @@
...
   <sup>[link](#dont-abuse-zero-key)</sup>
+  <details>
+    <summary><em>Explanation</em></summary>
+  
+    <pre>  
+      Pluralization rules vary from language to language and keys are automatically added and 
+      removed from the translation files. So, the `zero:` key cannot be relied on to be present 
+      for every language. 
+      
+      Use a specific key for the "no results" message instead.
```
davidstosik, line 269: This is just a bit confusing. (The "specific key" is not at the same level as `zero`.) Shall we include bad/good examples?

```yml
# Bad
search_results:
  zero: "There were no results"
  one: "1 recipe found"
  other: "%{count} recipes found"

# Good
search_results:
  one: "1 recipe found"
  other: "%{count} recipes found"
search_no_results: "There were no results"
```

rails/README.md
```diff
@@ -256,8 +256,19 @@
   <sup>[link](#scope-generic-phrases-under-common)</sup>
 
 - <a name="dont-abuse-zero-key"></a>
-  Don't use i18n `zero:` key to display a "no results" message (it is intended only to allow proper grammar)
+  Don't use i18n `zero:` key to display a "no results" message.
```
davidstosik, line 259:
```suggestion
  Don't use i18n `zero:` key to display a "no results" message.
```

jesseclark:
@davidstosik Thanks for the suggestion! The concrete example is a big improvement. ♻️

ChrisBr:
Here is the original explanation

> This is a very common misuse of pluralization; the main misconception
being that we can use the value of the zero pluralization to display a
special message in our app. This is misleading for translators who might
not even have pluralizations in their locale (Chinese, Japanese etc…). A
no_results translation is always clearer and doesn't need to infer
meaning from a numeric value.

https://github.com/cookpad/global-style-guides/pull/3/files

balvig:
👏 👍

PR #167, opened 2020-08-18
Add service object naming advice

samrjenkins:
@sikachu picked me up on this point of styling around the naming of service objects (https://github.com/cookpad/global-web/pull/20347#discussion_r471959054). Thought it would be worth considering adding to the guide

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -310,6 +310,24 @@
     ```
   </details>
 
+- <a name="service-object-naming"></a>
+  Prefer `ActorAction` format for naming service objects.
+  <sup>[link](#service-object-naming)</sup>
+
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```ruby
+    ## Bad
+    class UnpublishRecipeService
+    end
+
+    ## Good
+    class RecipeUnpublishService
+    end
+    ```
+  </details>
+
 - <a name="service-object-method"></a>
   Service objects should have a single public method `#run`, which does not accept arguments
   <sup>[link](#service-object-method)</sup>


balvig:
Looking at the current state of https://github.com/cookpad/global-web/tree/master/app/services, it looks like it's currently a bit more of a "free for all" 😓 

@samrjenkins would it possible to pick a few of the ones that don't follow this pattern and share in this PR as examples as how they would look?

I wonder about things like

```
      translate_base_service.rb                                                                                         
      translate_collection_service.rb                                                                                   
      translate_hash_service.rb                                                                                         
      translate_record_service.rb                                                                                       
      translate_service.rb  
```

for example

or

```
new_author_retention_emails_service.rb 
```

or

```
signup_service.rb 
```

In fact there are many different kinds of deviations, so would be good if could get just a little bit clearer picture of how this would affect our current codebase before 👍ing 🙇  (consistency sounds like it would be a positive, but want to be sure that `ActorAction` is also expressive enough?)

jesseclark:
> In fact there are many different kinds of deviations, so would be good if could get just a little bit clearer picture of how this would affect our current codebase before 👍ing 🙇 (consistency sounds like it would be a positive, but want to be sure that ActorAction is also expressive enough?)

When I indicated my 👍 , I had focused in on "**Prefer** `ActorAction` format for naming service objects" and figured that this wording still left us with some leeway for exceptions.

For things like:
> ```
>       translate_base_service.rb                                                                                         
>       translate_collection_service.rb                                                                                   
>       translate_hash_service.rb                                                                                         
>       translate_record_service.rb                                                                                       
>       translate_service.rb  
> ```

Another option could be to namespace them like:
```
Translation::BaseService
Translation::CollectionTranslationService
Translation::HashTranslationService
Translation::RecordTranslationService
Translation::TextTranslationService 
```

This would keep the related files together in a namespaced subdirectory in services and allow for the individual classes to follow the `ActorAction` format.

> ```
> signup_service.rb 
> ```

This could become `account_creation_service.rb` or `user_registration_service.rb`.

karlentwistle:
As the author of all the translations service I feel like I should say I'd be happy to change them as @jesseclark has suggested to live in a namespace 😄 Since the translation code was implemented in a piece meal fashion I could probably refactor some of these services to be PORO but I am not 100% sure on the future of one account / translations so I haven't done this yet since it might be ✂️ .

The other thing I wanted to flag is just looking at the filenames in the directory I also saw a pattern with `recipe` and `comment`

Screenshot 2020-08-19 at 17 17 50

Screenshot 2020-08-19 at 17 17 42

So it would be interesting how those would fit into the namespace recommendation. For example they could also live in a namespace `Recipe` and `Comment` but then we'd up with names like `Recipe::RecipeDeletionService` which feels a bit odd.

I totally agree it would be good to come up with some guidelines for naming the services though 👌 It might just be a bit tricker than "ActorAction" due to some corner cases.

tejanium:
Just to remind everyone this is a public thread

jesseclark:
> As the original author of all the translation services I feel like I should say I'd be happy to change them as @jesseclark has suggested to live in a namespace. Since the translation code was implemented in a piece meal fashion I could probably refactor some of these services to be a PORO but I am not 100% sure on the future of one account translations so I haven't done this yet since it might be ✂️ .

👍 It wasn't my intention to single out those services as being "wrong" or anything (as we haven't had any recommendations for naming up till now and the way they are named currently does provide a clear relationship between them) but rather to carry-on with @balvig's question around how those examples might fit with the "ActorAction" recommendation. Hope it didn't come across like that 😅 .

> So it would be interesting how those would fit in with the namespace recommendation. For example they could also live in a namespace `Recipe` and `Comment` but then we would end up with names like `Recipe::RecipeDeletionService` which feels a bit odd. I guess as @jesseclark says the wording is
> 
> > "**Prefer** ActorAction format for naming service objects" and figured that this wording still left us with some leeway for exceptions."
> 
> So maybe since these comment and recipe services already follow the convention they've earned the right to be in `app/services/*.rb` and dont need to be swept away into a subdirectory 😄
> 
> Or maybe we want to revisit the recommendation knowing we want to place these services into a namespace. For example `Recipe::DeletionService` and `Comment::DeletionService` feels pretty natural to me.

I don't think we need to sweep everything into namespaces but that it may make sense in some cases (and help organize a quite large directory).

However, I don't mind the bit of repetition in `Recipe::RecipeDeletionService` but `Recipe::DeletionService` seems fine to me too. 🤷‍♂️ 

Another thing that came to mind seeing some of the examples you cited (`comment_activation_service`, `recipe_bookmarking_service` etc.), is that the "Action" part of the recommendation seems most natural when it is using the noun for the event associated with the action (`activation`) or the present-participle form of the verb (`bookmarking`).

So, in the example of services related to recipe publication, `RecipePublicationService` and `RecipeUnpublicationService` seem preferable to me to `RecipePublishService` and `RecipeUnpublishService`.

This may start getting a bit complicated to capture in a concise recommendation though 😬 😅 

> I totally agree it would be good to come up with some guidelines for naming the services though 👌 It might just be a bit tricker than "ActorAction" due to some corner cases.

Agreed! 👍

karlentwistle:
> 👍 It wasn't my intention to single out those services as being "wrong" or anything (as we haven't had any recommendations for naming up till now and the way they are named currently does provide a clear relationship between them) but rather to carry-on with @balvig's question around how those examples might fit with the "ActorAction" recommendation. Hope it didn't come across like that 😅 .

It's all good 👌 They're a good example of current service names that don't fit the new mould so make for an interesting outlier to discuss.

balvig:
@samrjenkins @jesseclark Ah, I understand, but I guess "prefer" suggests that it _is preferable_ most of the time? 😄 

I think it's worth taking a look at what we already have and getting a feel for how when _consistently_ applying the preferred approach might affect the current code base, _whether_ we even need to allow for exceptions, and if so, how often/why we'd need to do so?

The ones I highlighted were just a couple that stood out at a quick glance, but @samrjenkins would be really great if you could do a pass as well, and maybe as an exercise share a few examples you come across, how they might be either modified to comply with the new rule, or any reasons why they shouldn't follow the new guideline?

PR #171, opened 2020-10-28
Update Kafka styleguide with message consistency guidelines

karlentwistle:
Trying to capture some of the discussion that has recently taken place in https://github.com/cookpad/platform-team/issues/10#issuecomment-716637676 into a simple to understand style guide recommendation.

We have realised that the existing guidelines about [keeping messages small](https://github.com/cookpad/global-style-guides/tree/master/kafka#keep-small-messages) needs some clarifications when the topic the messages are being emitted onto is configured with compaction enabled.

### DIFF
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -17,6 +17,14 @@
     "title": "edited title"
   }
 }
+{
+  "type": "registered_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "registered title"
+    "description": "registered description"
+  }
+}
 
 // Good
 {
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -14,7 +14,7 @@
   "type": "registered_user_event",
   "key": "users_123",
   "body": {
-    "title": "registered title"
+    "title": "registered title",
     "description": "registered description"
   }
 }
@@ -29,6 +29,14 @@
 
 
 // Good
+{
+  "type": "registered_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "registered title",
+    "description": "registered description"
+  }
+}
 {
   "type": "edited_user_event",
   "key": "users_123",
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -5,6 +5,30 @@
 - <a name="keep-small-messages"></a>
   Keep messages as small as possible. Expose just enough data so messages make sense to other systems. Use simple RESTish payloads. <sup>[link](#keep-small-messages)</sup> <sup>[explanation](https://martinfowler.com/articles/microservices.html#SmartEndpointsAndDumbPipes)</sup>
 
+- <a name="message-consistency"></a>
+  Keep messages payloads consistant between messages that share the same key format, if the topic is configured to be compacted. <sup>[link](#message-consistency)</sup>
+
+```json
+// Bad
+{
+  "type": "edited_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "edited title"
+  }
+}
+
+// Good
+{
+  "type": "edited_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "edited title",
+    "description": "previous description"
+  }
+}
+```
+
 - <a name="events-name-past-tense"></a>
   Prefer `{PastTenseVerb}_{Entity}`format for naming the events, event names should be in past tense. <sup>[link](#real-world-events) [explanation](https://youtu.be/JzWJI8kW2kc?t=707)</sup>
 
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -8,6 +8,8 @@
 - <a name="message-consistency"></a>
   Keep message payloads consistent between messages that share the same key format, if the topic is configured to be compacted, and if consumers expect to replay the compacted events and get a full picture of a resource's current state. <sup>[link](#message-consistency)</sup>
 
+<details>
+  <summary><em>Example</em></summary>
 ```json
 // Bad
 {
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -48,7 +48,7 @@
   }
 }
 ```
-
+</details>
 - <a name="events-name-past-tense"></a>
   Prefer `{PastTenseVerb}_{Entity}`format for naming the events, event names should be in past tense. <sup>[link](#real-world-events) [explanation](https://youtu.be/JzWJI8kW2kc?t=707)</sup>
 
--- a/kafka/README.md
+++ b/kafka/README.md
@@ -11,13 +11,6 @@
 ```json
 // Bad
 {
-  "type": "edited_user_event",
-  "key": "users_123",
-  "body": {
-    "title": "edited title"
-  }
-}
-{
   "type": "registered_user_event",
   "key": "users_123",
   "body": {
@@ -26,13 +19,22 @@
   }
 }
 
+{
+  "type": "edited_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "edited title"
+  }
+}
+
+
 // Good
 {
   "type": "edited_user_event",
   "key": "users_123",
   "body": {
     "title": "edited title",
-    "description": "previous description"
+    "description": "registered description"
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


kafka/README.md
```diff
@@ -5,6 +5,30 @@
...
+
+```json
+// Bad
+{
+  "type": "edited_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "edited title"
+  }
+}
```
robertomiranda, line 19: I think we need another example of an event with the same key
```suggestion
}
{
  "type": "registered_user_event",
  "key": "users_123",
  "body": {
    "title": "registered title"
    "description": "registered description"
  }
}
```

kafka/README.md
```diff
@@ -5,6 +5,30 @@
 - <a name="keep-small-messages"></a>
   Keep messages as small as possible. Expose just enough data so messages make sense to other systems. Use simple RESTish payloads. <sup>[link](#keep-small-messages)</sup> <sup>[explanation](https://martinfowler.com/articles/microservices.html#SmartEndpointsAndDumbPipes)</sup>
 
+- <a name="message-consistency"></a>
+  Keep messages payloads consistant between messages that share the same key format, if the topic is configured to be compacted. <sup>[link](#message-consistency)</sup>
```
robertomiranda, line 9: This also applies when we are designing `Event-Carried State Transfer` pattern where we want to carry the whole aggregate, wonder if we can incorporate that part, even without compaction we could share the same aggregate root by design.

what about? 

```suggestion
  Keep messages payloads consistent between messages that share the same key format, if the topic is configured to be compacted or is designed to transfer an aggregate/resource state <sup>[link](#message-consistency)</sup>
```

kafka/README.md
```diff
@@ -5,6 +5,30 @@
 - <a name="keep-small-messages"></a>
   Keep messages as small as possible. Expose just enough data so messages make sense to other systems. Use simple RESTish payloads. <sup>[link](#keep-small-messages)</sup> <sup>[explanation](https://martinfowler.com/articles/microservices.html#SmartEndpointsAndDumbPipes)</sup>
```
ollieh-m, line 6: It may just be me, because I'd thought this rule meant only include what is strictly relevant to the event that's happened... but do we need clarification or a caveat here? Something like:
> Bear in mind that "just enough" will be different in different contexts depending, for example, on whether the topic is being compacted, and whether consumers rely on getting a full picture by replaying the compacted event stream.

kafka/README.md
ollieh-m: I feel like this would need some more explanation. The phrase "transfer an aggregate/resource state" doesn't mean much to me. 

Also, if a topic has been designed to communicate the full resource state in each event, won't "message consistency" happen naturally? It's only really a rule we'd need to bear in mind if we are otherwise trying to streamline messages.

ollieh-m, line 9:
```suggestion
  Keep message payloads consistent between messages that share the same key format, if the topic is configured to be compacted, and if consumers expect to replay the compacted events and get a full picture of a resource's current state. <sup>[link](#message-consistency)</sup>
```
I've added another caveat in the rule, because I don't think message consistency is a concern unless we expect or need replayability, and it seems [replayability isn't always expected](https://github.com/cookpad/platform-team/issues/10#issuecomment-718046851). However, I'm not really sure whether we're aiming as a general principle for all consumers to be able to recreate their state by replaying the stream... If that is a principle we aim to follow, this extra caveat is irrelevant.

karlentwistle: @ollieh-m makes sense to me 

>  However, I'm not really sure whether we're aiming as a general principle for all consumers to be able to recreate their state by replaying the stream

No, I don't think we're aiming for that. Sometimes it makes sense to be able to recreate the state and other times it won't.

karlentwistle:
@robertomiranda I applied @ollieh-m suggestion so I wondered if you could ♻️ 😄

kafka/README.md
```diff
@@ -5,6 +5,48 @@
...
 
+- <a name="message-consistency"></a>
+  Keep message payloads consistent between messages that share the same key format, if the topic is configured to be compacted, and if consumers expect to replay the compacted events and get a full picture of a resource's current state. <sup>[link](#message-consistency)</sup>
```
jesseclark, line 9: It seems like this recommendation requires knowing the behavior of current consumers and potentially predicting future behavior of existing or new consumers. Would it be simpler and safer to make this recommendation unconditional?

```suggestion
  Keep message payloads consistent between messages that share the same key format. <sup>[link](#message-consistency)</sup>
```

jesseclark (approved):

kafka/README.md
```diff
@@ -5,6 +5,48 @@
...
+
+```json
+// Bad
```
robertomiranda, line 12: [nit] what do you think about wrapping using \<details\> tag?

<details>
  <summary><em>Example</em></summary>
  Hello Worlds
</details>

```suggestion
<details>
  <summary><em>Example</em></summary>
```json
// Bad
```

robertomiranda (approved):
Looks good 👌

kafka/README.md
karlentwistle: @jesseclark I applied that part based on a suggestion from @ollieh-m https://github.com/cookpad/global-style-guides/pull/171#discussion_r514118398. 

I think decisions around if a topic should be compacted or not are ones that should be taken when the topic is being designed (ideally before messages are being produced). Also depending on what the topic is being used for would influence if you wanted to compaction applied or not. Which would then influence how you'd design the messages.

jesseclark: I see. Thanks for clarifying. 

So, if a topic isn't going to be compacted is it safe/allowable to have inconsistent payloads under the same key format?

karlentwistle: > So, if a topic isn't going to be compacted is it safe/allowable to have inconsistent payloads under the same key format?

Yes, that's my understanding.

kafka/README.md
```diff
@@ -5,6 +5,50 @@
...
+{
+  "type": "edited_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "edited title",
+    "description": "registered description"
+  }
+}
+```
+
```
artygus, line 51: markup looks a bit broken under example section, I think it misses closing `</details>` tag

https://github.com/cookpad/global-style-guides/blob/467adfc82507a76c7e8efcf8e9ea610ec65018a4/kafka/README.md#topic-message-key

Screen Shot 2020-10-29 at 4 03 25 PM


```suggestion
</details>
```

kafka/README.md
```diff
@@ -5,6 +5,50 @@
...
+// Good
+{
+  "type": "registered_user_event",
+  "key": "users_123",
+  "body": {
+    "title": "registered title",
+    "description": "registered description"
+  }
+}
+{
```
artygus, line 42: [nit] should we add an empty line between messages like in a `// Bad` section?

```suggestion

{
```

kafka/README.md
```diff
@@ -5,6 +5,50 @@
 - <a name="keep-small-messages"></a>
   Keep messages as small as possible. Expose just enough data so messages make sense to other systems. Use simple RESTish payloads. <sup>[link](#keep-small-messages)</sup> <sup>[explanation](https://martinfowler.com/articles/microservices.html#SmartEndpointsAndDumbPipes)</sup>
 
+- <a name="message-consistency"></a>
+  Keep message payloads consistent between messages that share the same key format, if the topic is configured to be compacted, and if consumers expect to replay the compacted events and get a full picture of a resource's current state. <sup>[link](#message-consistency)</sup>
+
+<details>
+  <summary><em>Example</em></summary>
+```json
+// Bad
```
artygus, line 14: [nit] it took me a couple of seconds to find difference between good and bad, maybe we could expand the comment?

```suggestion
// Bad (notice missing description key in second message)
```

artygus (approved):
looks great overall! just a few suggestions on markup 🥣

ollieh-m (approved):

balvig (approved):
Great stuff 👏

kafka/README.md
dannyd4315: Agree with @ollieh-m here, it sounds like out thinking about the payload we are sending in events has evolved somewhat.

dannyd4315 (approved):

PR #172, opened 2020-11-19
Avoid adding translation for single language experiments

brunobrgs:
Follow up: https://github.com/cookpad/global-web/pull/21917

> Translations that have been added to the yaml files notifies the translation team who will attempt to add a 
    translation for the feature, if it is experimental and only needed for one language there is no need to add 
    your translation to the yaml file.

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -259,7 +259,7 @@
     <summary><em>Explanation</em></summary>
     
     Translations that have been added to the yaml files notifies the translation team who will attempt to add a 
-    translation for the feature, if it is experimental and only needed for one language there is no need to add 
+    translation for the feature. If it is experimental and only needed for one language there is no need to add 
     your translation to the yaml file.
   </details>
   
@@ -268,10 +268,10 @@
     
     ```ruby
     # Bad
-    t("key.present.in.yml")
+    t("new.key.in.yml")
 
     # Good
-    t("key.absent.in.yml", default: "Text in the country language")
+    t("absent.key.in.yml", default: "Text in the country language")
     ```
   </details>
 
--- a/rails/README.md
+++ b/rails/README.md
@@ -251,6 +251,30 @@
   Use fully qualified i18n name. eg. `recipes.show.title` rather than shorthand `.title`
   <sup>[link](#use-fully-qualified)</sup>
 
+- <a name="avoid-single-language-translation"></a>
+  Avoid adding i18n during experiments running for a single language
+  <sup>[link](#avoid-single-language-translation)</sup>
+  
+  <details>
+    <summary><em>Explanation</em></summary>
+    
+    Translations that have been added to the yaml files notifies the translation team who will attempt to add a 
+    translation for the feature, if it is experimental and only needed for one language there is no need to add 
+    your translation to the yaml file.
+  </details>
+  
+  <details>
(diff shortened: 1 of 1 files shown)


rails/README.md
```diff
@@ -251,6 +251,30 @@
...
+- <a name="avoid-single-language-translation"></a>
+  Avoid adding i18n during experiments running for a single language
+  <sup>[link](#avoid-single-language-translation)</sup>
+  
+  <details>
+    <summary><em>Explanation</em></summary>
+    
+    Translations that have been added to the yaml files notifies the translation team who will attempt to add a 
+    translation for the feature, if it is experimental and only needed for one language there is no need to add 
+    your translation to the yaml file.
```
balvig, line 263:
```suggestion
    translation for the feature. If it is experimental and only needed for one language there is no need to add 
    your translation to the yaml file.
```

balvig (approved):
👏

rails/README.md
```diff
@@ -251,6 +251,30 @@
...
+  
+  <details>
+    <summary><em>Example</em></summary>
+    
+    ```ruby
+    # Bad
+    t("key.present.in.yml")
+
+    # Good
+    t("key.absent.in.yml", default: "Text in the country language")
```
artygus, line 274: a minor wording thing, I think it makes intention a bit clearer  

```suggestion
    # Bad
    t("new.key.in.yml")

    # Good
    t("absent.key.in.yml", default: "Text in the country language")
```

artygus (approved):

davidstosik (approved):
Nice. 👍 

Thinking out loud: would there be any value in having `experiments.XX.yml` files, that would not go into OneSky but allow us to centrally manage experiment translations without bothering translators?

At least we'd be able to keep using `t(...)` in specs without having to provide the same default string in multiple places. (See [Use I18n helpers when clicking/asserting](https://github.com/cookpad/global-style-guides/tree/master/rspec#use-i18n).)

artygus:
> Thinking out loud: would there be any value in having experiments.XX.yml files, that would not go into OneSky but allow us to centrally manage experiment translations without bothering translators?

great option, covers the case when experiment might be extended to second language

alinagrishchuk (approved):

jesseclark:
> Thinking out loud: would there be any value in having `experiments.XX.yml` files, that would not go into OneSky but allow us to centrally manage experiment translations without bothering translators?

This seems preferable to me over the `default:` approach and would also enable us to share the file with the translators for the language that we need translated for the experiment. The keys could then be moved from the `experiments.en.yml` to the main translation file when the experiment was going to be rolled out.

dannyd4315:
> > Thinking out loud: would there be any value in having `experiments.XX.yml` files, that would not go into OneSky but allow us to centrally manage experiment translations without bothering translators?
> 
> This seems preferable to me over the `default:` approach and would also enable us to share the file with the translators for the language that we need translated for the experiment. The keys could then be moved from the `experiments.en.yml` to the main translation file when the experiment was going to be rolled out.

Yeah I really like this idea as well 👍

balvig:
> Thinking out loud: would there be any value in having `experiments.XX.yml` files\

Nice, this sounds like a genuine silver bullet to me 👏

PR #176, opened 2021-01-07
Add guideline to use prettier to format javascript code

tony-rowan:
## What
Update style guides to reflect our [adoption of Prettier](https://github.com/cookpad/web-chapter/issues/584).

## Why
Our style guides should be kept up to date with current our practices, to better help onboard new starters and provide context for discussions.

PR #177, opened 2021-01-21
 Add guideline to use prettier to format css

tony-rowan:
## What
Update the styleguides to reflect that we use Prettier to format CSS.

## Why
To avoid [surprising developers](https://github.com/cookpad/global-style-guides/pull/176#issuecomment-764734398) when inconsistent CSS is flagged. In addition to the reasons we [documentated](https://github.com/cookpad/global-style-guides/pull/176#issue-551016396) our usage of Pretteir for JS code.

> Our style guides should be kept up to date with current our practices, to better help onboard new starters and provide context for discussions.


alinagrishchuk (approved):

tejanium (approved):
🚀

PR #178, opened 2021-02-09
Update Python style guide

radeklat:
This PR:
* Updates the Python style guide to link to the new version
* Removes unused pylintrc config file for Pylint
* Remove the outdated Python style guide

The current Python style guide in this repository is outdated. The search team in collaboration with ML and MLI (the main users of Python in Cookpad) decided to create a new one from scratch, matching the current state. The new style guide has been stable for the last couple of month and these teams consider it de facto the style guide for Python.

The new Python style guide is in a separate repository to allow and encourage fast iteration, ownership and collaboration of the teams that use it. So we believe it should stay that way. There are couple of options of what to do with the old `cookpad/global-style-guides` for Python. I suggest linking to the `cookpad/python-style-guide` (this PR). However note that while `cookpad/global-style-guides` is public, `cookpad/python-style-guide` is private. So the link would work only for Cookpad employees. If that is a problem, we can either make `cookpad/python-style-guide` public or remove the link from the old style guide.


README.md
```diff
@@ -2,7 +2,7 @@
 
 - [Git](/git)
 - [Javascript](/javascript)
-- [Python](/python)
+- [Python](https://github.com/cookpad/python-style-guide)
```
ettiee, line 5: This repository is public whereas https://github.com/cookpad/python-style-guide is private so this link will be broken for non-employees. Can we make the python-style-guide public? If not then we should remove this link

radeklat: Literally what the PR description says :D Was it TL;DR or did you want to mention it specifically in here? I don't mind either way but since I don't own this repo, it doesn't feel like I should decide.

ettiee: Just confirming that IMO we shouldn't leave a broken link in a public repository. @karlentwistle wdyt?

karlentwistle: Yeah I think we want to avoid broken links in public repository

mattjw: @karlentwistle does it then make sense to remove python from `global-style-guides` and `global-style-guides` becomes `global-web-style-guides`?

AFAIK android and iOS maintain their own style conventions outside of this repo. Alternatively they could centralise them here, but they'd also want to check they don't have any internal links before centralising.

radeklat: @karlentwistle :point_up: any decision on this one?

karlentwistle: > @karlentwistle does it then make sense to remove python from global-style-guides and global-style-guides becomes global-web-style-guides?

Yep, I think this makes sense 👌

radeklat:
I don't know why it took me so long to remove the one link link to the Python guide but now it is done. @karlentwistle and/or @mattjw  can you please approve?

mattjw (approved):
LGTM

PR #185, opened 2021-03-04
Restore SCSS SelectorDepth max to 4

davidstosik:
## What

This sets the maximum nesting of selectors to 4, as it [used to be](https://github.com/cookpad/global-web/pull/4433/files#diff-ebef2f6706bbb40d0612571ca7f67fdf84ad7666e525d0572a35a090b7afe9a9R160).

## Why

You'll find a more complete explanation in my previous PR: [Remove unnecessary defaults from .scss-lint.yml](https://github.com/cookpad/global-style-guides/pull/184).
In brief, we used to have this tweaked to `4` but it appears that setting was lost when we [brought new defaults from upstream](https://github.com/cookpad/global-web/pull/4433).

## Anything else?

It looks like we've been ignoring the error thrown to us by Hound/Pronto for a while:

image

This change reduces the number of warnings for this linter from 323 to 42.

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -21,7 +21,7 @@ linters:
     enabled: false
 
   SelectorDepth:
-    max_depth: 2
+    max_depth: 4
 
   SelectorFormat:
     convention: hyphenated_BEM


artygus (approved):

dannyd4315 (approved):

tony-rowan (approved):

hahmed (approved):

lollkebear (approved):

balvig (commented):
@davidstosik do we have some examples of things that got caught on this but shouldn't have?

I might be misunderstanding this rule but, whether it was configured by mistake or not 😅, my _impression_ is actually that with the current way we write CSS, most of the time there shouldn't be a need for deep selectors. 

At most we will have a modifier wrapping an element, ie:


```css
.nav {
}

.nav__item {
}

.nav--active .nav__item {
}
```

(I realize this is for SCSS as well, which we don't really use anymore either, so maybe doesn't matter too much either way!)

davidstosik:
@balvig	Ah, I knew I was forgetting something! Thanks for reminding me.
You are right that the problem does not come up often anymore (check the dates):


In recent examples, we have:

```scss
// https://github.com/cookpad/global-web/pull/23852/files/d17844426b33fcc40d36e1b64f57a9c6d155de38#r585752115
.bottom-navigation {
  // ...

  .nav-link--active {
    //...

    .nav-link__avatar { ⬅️
```

```scss
https://github.com/cookpad/global-web/pull/20771#discussion_r492116580
.dropdown-menu {
  li > a.hidden { ⬅️
    // ...
```

davidstosik:
@balvig Looks like we'll be happy to stick to a `max_depth` of 2? Or should find a middle ground and use the [default 3](https://github.com/sds/scss-lint/blob/037c810aa0cf17d19062473ccde6a8b4f1e10277/config/default.yml#L164)?
I'm happy with either, and if we stick to 2, I'll close this PR.

jesseclark:
FWIW, the 3 levels of nesting in the bottom nav have now been removed: https://github.com/cookpad/global-web/commit/425fb800e3e98b4241ff1fe3a8ad3e6d925058ed

balvig:
@davidstosik my _impression_ is that 2 should be enough for how we currently write CSS so maybe we can try sticking with that?

davidstosik:
Roger.
Hopefully we can get back to this conversation in the future if we need to know how the decision happened. 🤞🏻 

@cookpad/web-chapter Please note that this means the related linter warnings are relevant and should be handled.

As a rule of thumb, all linter warnings that you encounter should be treated as relevant.

If you happen to notice other linter rules that you think are irrelevant, please do not leave them unanswered: either start a conversation right in the PR, or open an issue/RFC/PR to discuss and figure out what to do with it. 🙏🏻

PR #187, opened 2021-05-11
[RFC] Disable Rails/ContentTag

jesseclark:
### What

Disable the Rails/ContentTag cop.

### Why

There is an [open PR](https://github.com/rubocop/rubocop-rails/issues/260) for Rubocop that makes a pretty compelling argument that this cop targets the wrong method.

And you can see in the Rails docs that there is a legacy syntax section for the [#tag method](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-tag) but no deprecation or legacy syntax is mentioned for [#content_tag](https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag).

cc: @cookpad/web-chapter

### DIFF
--- a/.rubocop.rails.yml
+++ b/.rubocop.rails.yml
@@ -22,3 +22,6 @@ Rails/RequestReferer:
 
 Rails/DynamicFindBy:
   Enabled: false
+
+Rails/ContentTag:
+  Enabled: false


balvig:
Interesting find! 

I'm not sure about this - in particular the "targeting the wrong method" bit 🤔 

I recall `tag.something` being introduced as "the new hotness" back in Rails 5.1, to replace _both_ `content_tag(:something)` and `tag(:something)`:

```rb
# To produce <span class="bookmark"></span> before:
content_tag :span, nil, class: 'bookmark'
# After (because we will keep an index of tags that need closing, like <span>, vs those that do not, like <br>)
tag.span class: 'bookmark'

# To produce <div id="post_1"> before:
content_tag :div, post.title, id: dom_id(post)  
# After:
tag.div post.title, id: dom_id(post)

# To produce <br> before:
tag :br, nil, true
# After (becomes like span, we know this tag doesn't need closing)
tag.br

# Would be nice to support nesting too, so to produce <div id="header"><span>hello</span></div>, you'd:
tag.div(id: 'header') { |tag| tag.span 'hello' }
```

- https://github.com/rails/rails/issues/25195
- https://github.com/rails/rails/pull/25543

I certainly prefer the _syntax_ of `tag.something`.

The slowness is a little concerning, but I suppose we don't use Rails or Ruby mainly for its speed and ultimately _any_ Rails HTML helper will always be slower than a raw string.

I'm not good enough with numbers to tell from the benchmarks what using `tag.something` means in terms of total rendering speed on an "average page" in Real Life™️, whether for the kind of pages we build it constitutes enough of a performance problem that we should avoid it entirely?

However, the general approach we've taken so far is to use any syntactic sugar that's available in the framework and only drop down when/if needed (for hotspots that would usually mean going all the way to writing "manual" HTML).

So, not really concluding anything here 😅, but while it's true that it's hard to find any mention of what the future holds for `content_tag`, it _seems_ like the intent was for it to be replaced by `tag.something` and if the performance aspect isn't clearly a _major_ problem, it might still be worth encouraging the nicer syntax through linting?

davidstosik:
Yeah, I think the Rubocop cop might have been a bit misleading, and there are multiple levels to look at:

- the `tag(:div)` (no block, no content) was [announced](https://github.com/rails/rails/blob/a5cda0407fffe4214db3c40487d9d6394b391bc1/actionview/lib/action_view/helpers/tag_helper.rb#L203) as soon deprecated, in favor of `tag.div`
- `content_tag(:div` (with content passed either with a string or a block) is [commented](https://github.com/rails/rails/blob/a5cda0407fffe4214db3c40487d9d6394b391bc1/actionview/lib/action_view/helpers/tag_helper.rb#L261) as "legacy syntax"
- the new `tag.div` syntax might be relatively slower because of its use of `method_missing` but seems to be recommended above both the above (see the links)

About performance, it's always a bit tricky because we always want to keep things fast and optimized, but it's important to put it in relation to its cost in terms of development comfort.

In the linked issue's benchmark, there seems to be an average 2.6ms difference between the fastest and slowest of equivalent calls using different syntaxes. It's not difficult to see how that could add up if we use either a lot: if we have a page that makes 100 such calls in a single render, then that means we'd find a way to reduce render time by 260ms.

100 calls in a single render are not unrealistic for a page with no cache, but I would expect it to be a lot less when making good use of partial caching.

I want to also mention that I feel that we may have been abusing `tag` or `content_tag`, in situations where pure HTML would have been just fine, and should have been prefered above either.

```diff
-<%= tag :meta, content: "False", itemprop: "isAccessibleForFree" %>
+<meta content="False" itemprop="isAccessibleForFree">
```

```diff
-<%= content_tag :div, nil, class: "text-center sm:text-left card p-sm", data: { controller: "banner" } do %>
+<div class="text-center sm:text-left card p-sm" data-controller="banner">
```

Here's another data point: it looks like we currently have 120 of such calls to either `tag` or `content_tag` in our whole codebase (`ag -s ' (tag[. (]|content_tag)' app/{helpers,views} | wc -l`).


In the end, I agree that the performance issue _might_ be a concern, which we may benefit from looking into, but I also think that we should try to steer away from syntaxes that have been marked as "legacy" or "to be deprecated", and also feel less comfortable to use.

About the performance concern, would there be an easy way for us to profile calls to `tag` and `content_tag` using ScoutAPM?

jesseclark:
> content_tag(:div (with content passed either with a string or a block) is commented as "legacy syntax" (aaand, that comes from the same PR too... 🤦🏻‍♂️ )

I missed that line in the comments/description. I was scanning for a "Legacy syntax" section in the docs as they added for the old flavor of `tag`

On that note, what does "legacy syntax" even mean? Is this an official designation or does it just indicate "we like this other way better and maybe someday might deprecate these other ways"?

Either way, if there is possibly some performance gain to using `content_tag`, and it has not been deprecated, do we want Rubocop telling us we have to change it to the possibly slower new form of `tag` at this time?

Maybe if it does actually become deprecated we can address changing `content_tag` to `tag` at that point?

> I'm not sure either it would make sense to keep so many ways of doing the same thing. 🤔

Isn't "many ways of doing the same thing" kinda part of the core philosophy or Ruby? 😉 😄

davidstosik:
> Isn't "many ways of doing the same thing" kinda part of the core philosophy or Ruby? 😉 😄

I was expecting that comment, thank you for not letting me down! 😄 
Yeah, I guess you're right, but in a way, our style guides (and Rubocop automation too) are here in order to add a bit of consistency to the codebase. 🤔 

I agree though that the "legacy" state of both `content_tag` and `tag`'s old syntax is dubious at best. Not getting a single deprecation warning in 5 years after declaring "legacy" makes me think that they might be here to stay... 🤔

jesseclark:
> I was expecting that comment, thank you for not letting me down! 😄

Happy to have fulfilled the expected trolling function! 😆 

> Yeah, I guess you're right, but in a way, our style guides (and Rubocop automation too) are here in order to add a bit of consistency to the codebase. 🤔

I totally agree on consistency in general especially with regards to higher-level patterns and problematic syntax. 

In this instance, forcing consistency feels a bit low-level and unnecessary to me. However, if the team is in favor of requiring the new tag syntax (despite the possibly minor performance issues), I am fine with that. 

I just found it a bit annoying to have Rubocop ping me about an already existing usage of `content_tag` while I am extracting a partial when the usage doesn't really seem to have any significant impact.

balvig:
> found it a bit annoying to have Rubocop ping me about an already existing usage

Hehe, yes I know that feeling! 
I think that's maybe more of a general problem? 😓 

If you're just moving things around, it feels "unfair" to have a bot (or human!) tell you to fix old code that just happens to be inside the block you're moving. 😅 

(You're welcome to tell the bot "listen, I'm just moving this" in those cases btw!)

As for `content_tag :something` vs `tag.something`, although it's maybe not the biggest difference in the world, I guess all else being equal I _do_ like the latter syntax better and there's enough evidence to suggest that it might be The Future.

In that regard, our linter helps not only with consistency, but also in making us aware of alternatives (at any level) that, even if the difference is minor, we sort of accept are "better" (or at least not worth having to think about.)

I also agree with @davidstosik that a lot of the time, plain text HTML is perfectly adequate and even preferable! (otherwise we would be writing HAML 😅 )

Therefore, in the context of performance optimization, it would probably make more sense to go from `tag.something` to `<something>...</something>` rather than the intermediate step of moving to `content_tag :something`?

(incidentally, seems to me that if `method_missing` is a significant bottleneck, then it maybe wouldn't be unheard of to at least define static methods within Rails for some of the more common tags? 🤔 )

All that is to say that, I guess having talked this through, I'm personally leaning towards keeping the cop? 🤔

davidstosik:
> (incidentally, seems to me that if method_missing is a significant bottleneck, then it maybe wouldn't be unheard of to at least define static methods within Rails for some of the more common tags? 🤔 )

Great point!

I wanted to give it a try, and first thing I did was to write down a benchmark (basically the same that is in the PR @jesseclark linked), and that is when I realized the submitter's benchmark is using [`Benchmark.ms`](https://api.rubyonrails.org/v6.1.0/classes/Benchmark.html) (an ActiveSupport extension that "benchmarks realtime in milliseconds") not Ruby's method (which display time in seconds). 😅 

So it turns out that the values he showed are not seconds but milliseconds, and that the timings I mentioned in my previous comment must be divided by 1000! I thought my conclusions were a bit surprising...

So we're actually talking of a 1.3µs (microsecond, a millionth of a second) difference on average between a call to `content_tag` and a call to `tag.something`. We could have hundreds, if not thousands of such calls in a single page render without having a humanly noticeable impact on the rendering time.

jesseclark:
Ok. Thanks for the discussion, everyone. And @davidstosik thanks for the performance investigations!

The cop stays and I will close this ticket. 👍🏼

balvig:
Thank you for raising @jesseclark. 🙏 
Although we ended up leaving as is for now, I learned quite a bit from the discussion!👌

PR #188, opened 2021-05-24
Add prefer-custom-error-class-in-abstract-methods

jesseclark:
### What

Adds a section to the Ruby style guide recommending using a custom error class over a plain raise or raising NotImplementedError in abstract methods.

### Why

Per the outcome of this [RFC](https://github.com/cookpad/web-chapter/issues/688)

### DIFF
--- a/ruby/README.md
+++ b/ruby/README.md
@@ -322,3 +322,27 @@
     end
     ```
   </details>
+
+- <a name="prefer-custom-error-class-in-abstract-methods"></a>
+  Prefer a custom error class over a plain raise or raising NotImplementedError in abstract methods.
+  <sup>[link](#prefer-custom-error-class-in-abstract-methods)</sup>
+
+  <details>
+    <summary>Example</summary>
+
+    ```ruby
+    ## Bad
+    def feed_class
+      raise NotImplementedError
+    end
+
+    def search_source
+      raise "Not implemented yet"
+    end
+
+    ## Good
+    def feed_class
+      raise UnimplementedAbstractMethodError.new(self.class.name, __method__)
+    end
+    ```
+  </details>


tony-rowan (approved):

tejanium (approved):

alinagrishchuk (approved):

davidstosik (approved):
Thank you! 👏🏻 

What do you think of adding the Rubocop rule too?

sikachu (approved):

hahmed (approved):

dannyd4315 (approved):

lollkebear (approved):

jesseclark:
> Thank you! 👏🏻
> 
> What do you think of adding the Rubocop rule too?

@davidstosik Looking at the [docs](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/InheritException) for that Rubocop rule again, it says:

> This cop looks for error classes inheriting from `Exception` and its standard library subclasses, excluding subclasses of `StandardError`.

and the list of "illegal classes" to inherit from is:

```ruby
ILLEGAL_CLASSES =

    %w[
      Exception
      SystemStackError
      NoMemoryError
      SecurityError
      NotImplementedError
      LoadError
      SyntaxError
      ScriptError
      Interrupt
      SignalException
      SystemExit
    ].freeze
```

So it seems that this cop would flag something like this:
```class UnimplementedAbstractMethodError < NotImplementedError```

but not: `raise NotImplementedError` which is what we want to discourage.

Also, I checked to make sure that this cop was not explicitly disabled in our rubocop config and tested locally and this cop is enabled by default and will flag the inheritance example above. 

We could possibly change the enforced style to `standard_error` instead of `runtime_error` if that is what our preference would be, but I do not have a strong opinion on that.

davidstosik:
@jesseclark You are right, I was mistaken, that cop acts on inheritance, not `raise`. And there does not seem to be a cop, at the moment, that would achieve what I'm suggesting. Thanks again for the doc addition! 👍🏻

PR #189, opened 2021-05-28
Don't use i18n pluralization for different texts

brunobrgs:
## What

Adds explanation about how to use Rails pluralization, considering that the only key present in all languages is `other`

## Why

Details here: https://github.com/cookpad/web-chapter/issues/697

### DIFF
--- a/rails/README.md
+++ b/rails/README.md
@@ -295,6 +295,36 @@
     ```
   </details>
 
+- <a name="pluralization-different-texts"></a>
+  Don't use i18n pluralization for different texts.
+  <sup>[link](#pluralization-different-texts)</sup>
+  <details>
+    <summary><em>Example</em></summary>
+
+    ```yml
+    # The only key present in all languages is `other`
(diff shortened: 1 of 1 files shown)


rails/README.md
```diff
@@ -295,6 +295,36 @@
     ```
   </details>
 
+- <a name="pluralization-different-texts"></a>
+  Don't use i18n pluralization for different texts.
```
aqeelvn, line 299:
```suggestion
  Don't use i18n pluralization for controlling application logic.
```

aqeelvn (approved):

tony-rowan (approved):

balvig (commented):
Nice one @brunobrgs, this looks good to me, but I'm wondering though if maybe we could try to merge it with https://github.com/cookpad/global-style-guides/tree/main/rails#dont-abuse-zero-key, as I think it's essentially the same issue.

Maybe something like `Only use i18n pluralization to preserve correct grammar`, and then having both the example you added along with the `zero` one?

balvig (approved):
👏 👍

lollkebear (approved):

PR #190, opened 2022-03-30
Rephrase RSpec style guide rule on matchers

Bodacious:
## What 

Rephrase RSpec stye guide rule on built-in matchers.

New phrasing places emphasis on the matcher descriptiveness, and not whether they are built-in to RSpec or not. 

## Why 

Current phrasing is unclear. Suggests `be_truthy` and `be_falsey` should be avoided because they are not built-in matchers, but they are.

https://relishapp.com/rspec/rspec-expectations/v/3-8/docs/built-in-matchers

## How 

Rephrase rule.

### DIFF
--- a/rspec/README.md
+++ b/rspec/README.md
@@ -432,7 +432,7 @@ These are some of the conventions we follow:
     ```
   </details>
 - <a name="built-in-matcher"></a>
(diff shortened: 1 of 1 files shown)


ollieh-m (approved):
Clearer 👍. Thanks @Bodacious 🙇‍♂️.

PR #192, opened 2022-07-14
Do not enforce Hash shorthand syntax in Ruby

Bodacious:
## What 

This PR adds a new Rubocop directive to prevent automatic enforcement of the hash shorthand syntax in Ruby.

## Why

Ruby 3.1 has added a [new shorthand syntax for hashes](https://bugs.ruby-lang.org/issues/17292), similar to Javascript's object shorthand.

``` ruby 
##
# This 
def one
  1
end 

two = 2

my_hash = { one: one, two: two} # => { one: 1, two: 2 }

##
# Can now be stated as
def one
  1
end 

two = 2

my_hash = { one:, two:} # => { one: 1, two: 2 }
```
_(Note: The hash in the final line does not have to explicitly name the method and local variables, since they share the same name as the hash keys)._

The default behavior in newer versions of Rubocop is to enforce this style wherever possible. [See this commit](https://github.com/rubocop/rubocop/commit/d85a3c496e0077b2fafa4862937ff472015e094e).

I think we need to discuss as a team: 
1. Whether we want to adopt this new shorthand syntax or not? (Happy to discuss in this PR)
2. Do we want Rubocop to automatically update existing code to use this new syntax or not?

## How 

For now, I've disabled this syntax in the Rubocop config file, to ensure our current code still passes linting.

cc: @cookpad/global-web-devs-all

### DIFF
--- a/.rubocop.ruby.yml
+++ b/.rubocop.ruby.yml
@@ -297,3 +297,7 @@ Style/PercentLiteralDelimiters:
 
 Naming/MemoizedInstanceVariableName:
   EnforcedStyleForLeadingUnderscores: required
+
+Style/HashSyntax:
+  EnforcedShorthandSyntax: never
+  


jesseclark (approved):

tony-rowan (approved):

tejanium (approved):
I hope we will turn this on when we use 3.1. This is one of the new features that I can't wait to use.

Bodacious:
@tejanium I'm happy for us to activate the feature on 3.1. I think we _should_ embrace Ruby's new features (except [rigtward assignment](https://bugs.ruby-lang.org/issues/15921) 🤢)

Do you have any thoughts on how we should handle existing code that could be updated to use this shorthand?

alinagrishchuk (approved):
🎉

tejanium:
> Do you have any thoughts on how we should handle existing code that could be updated to use this shorthand?

Don't worry about it, someone on the internet will write a script (or a text editor plugin) to update old syntax to a new syntax like it was in 2012 https://thoughtbot.com/blog/convert-ruby-1-8-to-1-9-hash-syntax 🤣

tejanium:
> Do we want Rubocop to automatically update existing code to use this new syntax or not?
> rubocop -a will automatically update these

Ah, I see, but I don't think we currently running rubocop in `-a` mode. I believe it just acts as a linter. We can update the syntax as we touch the file that contains old syntax, so the noise on PR can be minimum.

Or have someone do a big bang update like we did it with Stimulus' syntax https://github.com/cookpad/global-web/pulls?q=is%3Apr+sort%3Aupdated-desc+stimulus+is%3Aclosed+author%3Alollkebear+%22apply+new+stimulus+syntax%22

(I have a feeling, we only have a handful of old syntax in our codebase that can be safely converted into a new one)

Bodacious:
> Or have someone do a big bang update

I think we should avoid a Big Bang update. But it would be good if we could support both syntaxes in the linter and let teams gradually update the code over time.

Bodacious:
> (I have a feeling, we only have a handful of old syntax in our codebase that can be safely converted into a new one)

@tejanium https://github.com/cookpad/global-web/pull/31573 😂

tejanium:
I told you it's a handful

multi-hands

PR #193, opened 2022-07-14
Sort Rubocop directives in alphabetical order

Bodacious:
This PR sorts the Rubocop Ruby configuration directives in alphabetical order.

I haven't changed/added/removed any of the rules.

I hope this doesn't seem pedantic, but I find it easier to read the rules and make sure there are no dupes when they are kept in order.

---

Code used to sort this file:

``` ruby
require "yaml"

filepath = "./.rubocop.ruby.yml"
hash = YAML.safe_load_file(filepath)
sorted_hash = Hash[hash.sort]
File.write(filepath, sorted_hash.to_yaml)
```

### DIFF
--- a/.rubocop.ruby.yml
+++ b/.rubocop.ruby.yml
@@ -1,27 +1,120 @@
 Bundler/OrderedGems:
   ConsiderPunctuation: true
 
-Metrics/BlockLength:
+Layout/ArgumentAlignment:
+  EnforcedStyle: with_fixed_indentation
+
+Layout/ConditionPosition:
+  Enabled: false
+
+Layout/DotPosition:
+  EnforcedStyle: trailing
+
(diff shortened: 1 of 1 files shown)


arunagw (approved):

tony-rowan (approved):

Knack (approved):

Knack:
What if we do the same with `.rubocop.rails.yml`? We could do it even manually; it's small 😄

Bodacious:
> What if we do the same with `.rubocop.rails.yml`? We could do it even manually; it's small 😄

I was waiting to see how this PR was received before doing that. But I think it's a good idea :)

Bodacious:
@Knack  PR for the Rails file #194 :)

PR #196, opened 2022-11-04
Introduce rspec rubocop

shinsukeimai:
# What
Introduces rspec rubocop 

# Why
In `global-web`, we disabled rspec rubocop. However, in some cases we would like to point out our coding rule violation. So I add rpsec rubocop.

https://github.com/cookpad/global-web/pull/33185

# How
I just added one rule as a first step.
https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/DescribedClass

https://github.com/cookpad/global-style-guides/blob/028a31d62de5ce056cd3801080ee36e9285b5b90/.rubocop.rspec.yml#L51-L52


tejanium (approved):
Does rubocop have the option to disable all and then enable only what you need one by one?

Bodacious:
Does rubocop have the option to disable all and then enable only what you need one by one?

https://docs.rubocop.org/rubocop/configuration.html#generic-configuration-parameters
`DisabledByDefault`

tejanium:
Ah nice, I thought it was never implemented https://github.com/rubocop/rubocop/issues/1918#issuecomment-269494243

shinsukeimai:
https://github.com/cookpad/global-style-guides/pull/196/commits/dd12839134841d8c7ccd59a739099cb3f7dbe204

I checked it works on global-web by fetching via inherit_from.
We can use this not globally if we import it via `inherit_from`.

PR #197, opened 2022-11-09
Avoid excessive numbered params in a single block

robertomiranda:
## What

Avoid excessive numbered params in a single block

```
# bad
foo { _1.call(_2, _3, _4) }

# good
foo { do_something(_1) }
```

## Why

Detects the use of an excessive amount of numbered parameters in a single block. Having too many numbered parameters can make code too cryptic and hard to read.

cc @cookpad/global-rails

### DIFF
--- a/.rubocop.ruby.yml
+++ b/.rubocop.ruby.yml
@@ -223,6 +223,10 @@ Style/Next:
 Style/NilComparison:
   Enabled: false
 
+Style/NumberedParameters:
+  Enabled: true
+  EnforcedStyle: allow_single_line
+
 Style/NumberedParametersLimit:
   Enabled: true
   Max: 1
--- a/.rubocop.ruby.yml
+++ b/.rubocop.ruby.yml
@@ -223,6 +223,10 @@ Style/Next:
 Style/NilComparison:
   Enabled: false
 
+Style/NumberedParametersLimit:
+  Enabled: true
+  Max: 1
+
 Style/NumericLiterals:
   Enabled: false
 


shinsukeimai (approved):

.rubocop.ruby.yml
```diff
@@ -223,6 +223,10 @@ Style/Next:
 Style/NilComparison:
   Enabled: false
 
+Style/NumberedParametersLimit:
+  Enabled: true
+  Max: 1
+
```
tony-rowan, line 229: Suggestion: also restrict to single line. [[docs](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/NumberedParameters)]


```suggestion
Style/NumberedParameters:
  Enabled: true
  EnforcedStyle: allow_single_line

Style/NumberedParametersLimit:
  Enabled: true
  Max: 1

```

tony-rowan (approved):

.rubocop.ruby.yml
jesseclark: +1 for the single-line restriction.

jesseclark (approved):

lucianghinda (approved):

.rubocop.ruby.yml
```diff
@@ -223,6 +223,14 @@ Style/Next:
 Style/NilComparison:
   Enabled: false
 
+Style/NumberedParameters:
+  Enabled: true
+  EnforcedStyle: allow_single_line
+
+Style/NumberedParametersLimit:
+  Enabled: true
+  Max: 1
```
Bodacious, line 232: 👌

Bodacious (approved):

jesseclark (approved):

Knack (approved):

.rubocop.ruby.yml
ollieh-m: Can I double-check my understanding: `Avoid excessive numbered params in a single block` means `Avoid more than 1 numbered param in a single block`?

PR #198, opened 2022-11-28
Enable Lint/AssignmentInCondition

Knack:
Let's enable (set the default value for) the cop "[Lint/AssignmentInCondition](https://docs.rubocop.org/rubocop/cops_lint.html#lintassignmentincondition)". It was probably disabled when we started using RuboCop to keep compatibility with the code at that time.

With the default value, it allows you to keep the assignment in the condition if you wrap it in parenthesis:

```
Lint/AssignmentInCondition: Use == if you meant to do a comparison or wrap the expression in parentheses to indicate you meant to assign in a condition.
```

### DIFF
--- a/.rubocop.ruby.yml
+++ b/.rubocop.ruby.yml
@@ -31,9 +31,6 @@ Lint/AmbiguousOperator:
 Lint/AmbiguousRegexpLiteral:
   Enabled: false
 
-Lint/AssignmentInCondition:
-  Enabled: false
-
 Lint/DeprecatedClassMethods:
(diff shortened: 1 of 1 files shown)


rikarumi (approved):
🙏🏼

ollieh-m (approved):

robertomiranda (approved):

tony-rowan (approved):

alinagrishchuk (commented):
Thank you! 🥳 👏

PR #200, opened 2022-11-29
Disable Conditional Safe Assignment

robertomiranda:
## What
Disable Conditional Safe Assignment

## Why

Assignments in conditional assignments can be confused with the boolean operation, for the code author and reviewers.

## How

Disabling Safe assignments in conditionals we can improve the readability of the code by keeping assignments and conditions in separate lines.

## Anything else

Since Ruby 2.0 this type of conditional statement raise warnings in the console https://github.com/ruby/ruby/commit/e672994d146

cc @cookpad/global-rails

### DIFF
--- a/.rubocop.ruby.yml
+++ b/.rubocop.ruby.yml
@@ -26,7 +26,7 @@ Layout/LineLength:
   - spec/**/*
 
 Lint/AllowSafeAssignment:
-  Enabled: False
+  Enabled: false
 
 Lint/AmbiguousOperator:
   Enabled: false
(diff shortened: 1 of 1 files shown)


Bodacious (approved):

hardwinder (approved):
🚀

ainame (approved):
🦺

ollieh-m (approved):
I quite like assigning to a variable in a condition, but I see how a blanket rule against that is safer, so for the greater good ✅

jesseclark (approved):

.rubocop.ruby.yml
```diff
@@ -25,6 +25,9 @@ Layout/LineLength:
   - Gemfile
   - spec/**/*
 
+Lint/AllowSafeAssignment:
+  Enabled: False
```
robertomiranda, line 29:
```suggestion
  Enabled: false
```

PR #201, opened 2022-11-30
Fix AllowSafeAssigment

Knack:
`AllowSafeAssignment` is an option of `Lint/AssignmentInCondition`, not a cop by itself.


robertomiranda:
thanks @Knack 😅

Bodacious:
Thanks @Knack

